PR #1, opened 2016-02-15
Feature/scrutinizer

kallehauge:


kallehauge:
Don't merge :warning:

.eslintrc
```diff
@@ -0,0 +1,97 @@
...
+    "max-nested-callbacks": [1, 3],
+    "valid-jsdoc": [1, {
+      "prefer": {
+        "returns": "return",
+        "property": "prop"
+      },
+      "requireReturn": false
+    }]
+  }
+}
```
kasperg, line 97: Why do we need this file when we have [`web/.eslintrc`](https://github.com/DBCDK/biblo-admin/blob/master/web/.eslintrc)?

.scrutinizer.yml
```diff
@@ -0,0 +1,24 @@
...
+      - test
+      - theme
+    config:
+      standard: Drupal
+    filter:
+      paths: ['web/modules/custom/*', 'web/profiles/custom/*', 'web/themes/custom/*']
+  eslint:
+    use_native_config: true
+    filter:
+      paths: ['web/modules/custom/*', 'web/profiles/custom/*', 'web/themes/custom/*']
```
kasperg, line 18: Do we need this filtering when we do not have third party code in our repo?

kallehauge, line 18: Scrutinizer will build composer if there's a `composer.lock` file in the root

.eslintrc
kallehauge, line 97: Scrutinizer assumes that the `.eslintrc` file is in the root at the `.scrutinizer` file and I haven't found any ways to "point" to another file

.scrutinizer.yml
kasperg, line 18: :+1:

.eslintrc
kasperg, line 97: If they are equal then perhaps we could we note that somewhere?

kallehauge, line 97: True. I'll add a section to the `README.md` file about Scrutinizer in general :relaxed:

kasperg, line 97: Roger. When this is done I think this is ready to merge.

kasperg:
Great. Merged in 3714cb8!

giphy

PR #4, opened 2016-02-17
Add support for "secret" docker-compose env vars

kallehauge:
Ignore docker-compose.override.yml so we can provide secret docker environment variables.

Provide an example file to show what information that should be overridden.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@
 !/web/modules/custom
 !/web/themes/custom
 !/web/profiles/custom
+
+# Ignore Docker Compose override file to provide secret keys.
+docker-compose.override.yml
--- a/README.md
+++ b/README.md
@@ -14,8 +14,9 @@ The project structure is based on [`drupal-composer/drupal-project`](https://git
 1. Clone this repository
 2. From the root of the cloned repository:
   1. Run `composer install`
-	2. Run `docker-compose build`
-	3. Run `docker-compose up`
+  2. Copy `example.docker-compose.override.yml` to `docker-compose.override.yml` and fill out the required information.
+  3. Run `docker-compose build`
+  4. Run `docker-compose up`
 3. Profit!
 
 ## Webservice client generation
--- a/README.md
+++ b/README.md
@@ -14,10 +14,13 @@ The project structure is based on [`drupal-composer/drupal-project`](https://git
 1. Clone this repository
 2. From the root of the cloned repository:
   1. Run `composer install`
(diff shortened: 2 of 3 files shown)


example.docker-compose.override.yml
```diff
@@ -0,0 +1,5 @@
+service:
+  environment:
+    AMAZON_S3_KEYID: AmazonS3KeyId
+    AMAZON_S3_KEY: AmazonS3Key
+    UNILOGINSECRET: UniLoginSecret
```
kallehauge, line 5: @kasperg should "UNILOGINSECRET" be part of the override example?
I guess it shouldn't because it's only for talking between to your local instance?

kasperg, line 5: If it would otherwise be defined by restricted packages then I think it should.

README.md
```diff
@@ -14,8 +14,9 @@ The project structure is based on [`drupal-composer/drupal-project`](https://git
 1. Clone this repository
 2. From the root of the cloned repository:
   1. Run `composer install`
-	2. Run `docker-compose build`
-	3. Run `docker-compose up`
+  2. Copy `example.docker-compose.override.yml` to `docker-compose.override.yml` and fill out the required information.
```
kasperg: Should we treat this as a "required" step? The service app seems to work (as in it starts and you are able to perform some service calls) even with dummy values.

.gitignore
```diff
@@ -7,3 +7,6 @@
 !/web/modules/custom
 !/web/themes/custom
 !/web/profiles/custom
+
+# Ignore Docker Compose override file to provide secret keys.
+docker-compose.override.yml
```
kasperg, line 12: :+1:

kasperg:
Well done with the overrides. I have a comment regarding the documentation.

example.docker-compose.override.yml
kallehauge, line 5: Ah, I found the unilogin secret in the configuration:

> Sensitive information that should not be shared outside of the DBC A/S organization.

So this is part of the override.

kasperg, line 5: :+1:

README.md
kallehauge: That is a GREAT question! - But if we don't inform the reader about the information here, then we need it somewhere else.

Should we split up the installation into some sort of "Basic usage" and "Real usage" ?

kasperg: We could add two sections under Installation instructions: Required steps and Optional steps.

kallehauge:
@kasperg now with optional installation instructions

kasperg:
:+1:

PR #5, opened 2016-02-17
Update .gitignore

kallehauge:
Our exclude paths didn't work, so we will have to explicitly tell git what files/paths to ignore.


kasperg:
:+1:

PR #6, opened 2016-02-17
Add initial Drupal configuration

kallehauge:


kallehauge:
We could, at some point in time, consider making a "`drush cim -y`"  part of the Docker process. But it would require some workaround since we would have to make sure Docker is standing in `/web/` before running the command. – But I like the idea! :relaxed:

PR #7, opened 2016-02-17
Use adminimal as default theme.

kasperg:
Sooo pretty.


web/config/sync/core.extension.yml
```diff
@@ -6,6 +6,7 @@ module:
   dynamic_page_cache: 0
   field: 0
   filter: 0
+  help: 0
```
kallehauge, line 9: This is a bit out of the commits scope (enable module)

kallehauge:
When the "help" module is fixed (just a separate commit, not necessarily a new PR), then I'm satisfied and you can merge :relaxed:

web/config/sync/core.extension.yml
kasperg, line 9: [Adminimal depends on the help module](https://www.drupal.org/node/2629744) so enabling the help module is a requirement. I have updated the commit comment to explain this.

kallehauge:
:+1: I approve of this change

PR #9, opened 2016-02-22
#475: Moderator role

kasperg:
This pull request adds the Moderator role for users who will perform content and user moderation,

The role only has a few rudimentary permissions at the moment. More should be added later when moderation modules are implemented.

Also we only allow admins to create users. Moderators should not be able to sign up for themselves.


web/config/sync/user.role.moderator.yml
```diff
@@ -0,0 +1,14 @@
+uuid: d7b1ca0e-4271-4124-a560-9610e1a55679
+langcode: en
+status: true
+dependencies: {  }
+id: moderator
```
kallehauge: Should we use the id `moderator` ? I think it would be a good idea to prefix our moderator/editor roles with "admin-" since the CommunityService also contains roles (and even a moderator role?)

web/config/sync/user.settings.yml
```diff
@@ -9,7 +9,7 @@ notify:
   register_admin_created: true
   register_no_approval_required: true
   register_pending_approval: true
-register: visitors_admin_approval
+register: admin_only
```
kallehauge: :+1:

web/config/sync/user.role.moderator.yml
kallehauge: Hmm, we might have to rename the `administrator` role to keep it consistent, if we decide on this, even though there are no administrators in the service.

kasperg: In which cases do you think the overapping name would be a problem?

kasperg:
@kallehauge: I have prefiex role ids.

kallehauge:
@kasperg the labels and descriptions doesn't include any "Admin" prefix but I'm 50/50 if this matters. You can merge if you want :+1:

PR #10, opened 2016-02-22
Administrator role

kasperg:
This role will be given all permissions


kasperg:
Closed in favour of #11.

PR #14, opened 2016-02-24
Feature: Custom DBCDK Community module

kallehauge:
_**This PR includes the following:**_
Add a custom DBCDK Community Service module. This module contains a simple admin form that can be expanded later on, if we need for options/configuration.
It also contains an abstract "Community Block" class that will be the base for our future blocks containing different lists and other data gathered from the community service.


kallehauge:
I imagine that this module now contains the basics for what we need to really start building. - I have created an abstract class for our blocks, that currently only contains the API Client. This can ofc be expanded in the future if we find that our lists could benefit from some of the same methods etc. But I would rather extend on it from the beginning, so we don't have to "refactor" later on.

_**NB!**_ there is an error in the auto-generated client, so this has to be fixed before we can actually use the API (but the module itself is fine).

kasperg:
Lets talk about this face to face.

kallehauge:
I have defined the Community Services we need as services within Drupal, so if we want to use a service, now, we could do something like this (instead of creating a new instance of the service we want):

```
$post_api = \Drupal::service('dbcdk_community.api.post');
foreach ($post_api->postFind() as $post) {
  echo $post->getTitle();
}
```

I have created a configuration class (also registered as a service so it's easy to override) that extends on the Community Service Configuration that is passed as a dependency to the client in the `*.services.yml` file.

web/modules/custom/dbcdk_community/src/Form/SettingsForm.php
```diff
@@ -0,0 +1,70 @@
...
+    return ['dbcdk_community.settings'];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $config = $this->config('dbcdk_community.settings');
+
+    // Logo settings for theme override.
```
kasperg: Logo settings?

web/modules/custom/dbcdk_community/src/CommunityApiConfiguration.php
```diff
@@ -0,0 +1,35 @@
...
+   * Set configuration for the Community Service Api Client.
+   */
+  public function __construct() {
+    parent::__construct();
+
+    $config = \Drupal::config('dbcdk_community.settings');
+    // The Community Service prefixes each endpoint with a slash,
+    // so we make sure the host does not end with one.
+    $host = rtrim($config->get('community_service_url'), '/');
+    $this->setHost($host);
```
kasperg, line 32: We should consider also adding an option for setting [the debug configuration](https://github.com/DBCDK/biblo-admin/blob/6ac5aa10f1b797b6529a4c65306792e9a178e289/lib/client/src/Configuration.php#L391-L391). This is not necessarily needed right now but could be useful going forward.

kallehauge, line 32: Sure! What are you thinking about the "enable" flow? A setting you define in your settings.php file or a checkbox in the admin form?

kasperg, line 32: I have not thought this through at the moment. I just noticed that options were available and wanted to throw it out there.

kasperg:
@kallehauge: :+1: Looks good to me. Do you want to rebase before merging?

kallehauge:
@kasperg I rebased it last night and we haven't merged anything since, so it should be good to go. If you would confirm this by merging, then you'll get 1000x karma points :+1: :smile:

kasperg:
I meant rebase to squash some commits, but OK.

web/modules/custom/dbcdk_community/src/CommunityApiConfiguration.php
kallehauge, line 32: Fair enough ^.^ - By enabling it as a checkbox in the admin form (at some point in time) under an `advanced settings` tab, then it would still be possible to override it in settings.php if the developer wished to do so. – But we can talk about that when the need arises.

PR #15, opened 2016-02-25
Disable image scaling queue in community service

kasperg:
It requires a locally run Redis which is not available in our Docker
environment. It is not a required dependency so we use an environment
variable to disable it. This removes connection errors in the log.


kasperg:
This change is related to [this commit](https://github.com/DBCDK/dbc-community-service/commit/7a8971b093ef9905506c9e38ed55d032eff01bef#commitcomment-16321563).

kallehauge:
@kasperg: If you add the environment variable to the `example.docker-compose.override.yml` file, then you can merge at will :)
Nice spotted with the change :+1:

kasperg:
`example.docker-compose.override.yml` updated. I will merge then.

PR #16, opened 2016-02-29
Initial test setup

kasperg:
This PR defines an initial test setup including:
- Syntax checks (using php-parallel-lint)
- Unit tests (using PHPUnit) 
- Browser tests (using Behat)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,11 @@
         "drupal/console": "~0.10",
         "drupal/adminimal_theme": "^8.1"
     },
+    "require-dev": {
+        "behat/behat": "^3.0",
+        "drupal/drupal-extension": "^3.0",
+        "lakion/mink-debug-extension": "^1.0"
+    },
     "conflict": {
         "drupal/drupal": "*"
     },
@@ -28,7 +33,9 @@
     "prefer-stable": true,
     "scripts": {
         "post-install-cmd": "sh ./scripts/composer/post-install.sh",
-        "docker-swagger-generate": "docker-compose run swagger generate -i http://service:3000/explorer/swagger.json -l php -o /var/usr/client -c /var/usr/swagger/config.json"
+        "docker-swagger-generate": "docker-compose run swagger generate -i http://service:3000/explorer/swagger.json -l php -o /var/usr/client -c /var/usr/swagger/config.json",
+        "behat": "vendor/bin/behat -c tests/behat.yml",
+        "docker-behat": "docker-compose run web ../vendor/bin/behat -c ../tests/behat.yml --profile docker"
     },
     "extra": {
         "installer-paths": {
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,9 @@
         "behat/behat": "^3.0",
         "drupal/drupal-extension": "^3.0",
         "lakion/mink-debug-extension": "^1.0",
-        "jakub-onderka/php-parallel-lint": "^0.9.2"
+        "jakub-onderka/php-parallel-lint": "^0.9.2",
+        "mikey179/vfsStream": "~1.2",
+        "phpunit/phpunit": "~4.8"
     },
     "conflict": {
         "drupal/drupal": "*"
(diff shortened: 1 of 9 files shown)


composer.json
```diff
@@ -28,7 +37,11 @@
     "prefer-stable": true,
     "scripts": {
         "post-install-cmd": "sh ./scripts/composer/post-install.sh",
-        "docker-swagger-generate": "docker-compose run swagger generate -i http://service:3000/explorer/swagger.json -l php -o /var/usr/client -c /var/usr/swagger/config.json"
+        "docker-swagger-generate": "docker-compose run swagger generate -i http://service:3000/explorer/swagger.json -l php -o /var/usr/client -c /var/usr/swagger/config.json",
```
bicschneider, line 40: Should the swagger run as part of the pipeline?
Alternative: 
http://zircote.com/swagger-php/using_swagger.html#via-command-line

```
php swagger.phar /projects/my_project  -o /var/html/swagger-docs
```

kasperg, line 40: We cannot use Swagger-PHP. We use Swagger for generation of [webservice client code](https://github.com/DBCDK/biblo-admin/tree/develop/lib/client) from a Swagger specification. Swagger-PHP generates a Swagger specification from your webservice implementation.

My initial plan was not to generate client code as a part of the pipeline. The developer should do that as a part of the development process. If the specification changes and the generated code is updated during deployment then we may end up in a situation where the surrounding code calls a method that no longer exists. Such situations should be caught much earlier.

composer.json
```diff
@@ -21,6 +21,15 @@
         "drupal/console": "~0.10",
         "drupal/adminimal_theme": "^8.1"
     },
+    "require-dev": {
+        "behat/behat": "^3.0",
+        "drupal/drupal-extension": "^3.0",
+        "lakion/mink-debug-extension": "^1.0",
+        "jakub-onderka/php-parallel-lint": "^0.9.2",
+        "mikey179/vfsStream": "~1.2",
```
kallehauge, line 29: Is it because of the "image scale" service that you already include vfsStream as a dependency now, @kasperg ? :relaxed:

composer.json
```diff
@@ -28,7 +37,11 @@
     "prefer-stable": true,
     "scripts": {
         "post-install-cmd": "sh ./scripts/composer/post-install.sh",
-        "docker-swagger-generate": "docker-compose run swagger generate -i http://service:3000/explorer/swagger.json -l php -o /var/usr/client -c /var/usr/swagger/config.json"
+        "docker-swagger-generate": "docker-compose run swagger generate -i http://service:3000/explorer/swagger.json -l php -o /var/usr/client -c /var/usr/swagger/config.json",
+        "lint": "vendor/bin/parallel-lint . --exclude vendor --exclude web/core --exclude web/sites/*/files",
```
kallehauge, line 41: What does `php-parallel-lint` catch that PHPCS doesn't?

docker-compose.yml
```diff
@@ -62,3 +64,12 @@ swagger:
     - './lib/client:/var/usr/client'
   links:
     - service
+hub:
+  image: selenium/hub
+  ports:
+    - "4444"
+firefox:
```
kallehauge, line 77: This is more my curiosity than a real question to the PR: Why Firefox? :) Is there any advantages to the Firefox webdriver vs ex Chrome? – I know you use Chrome as your primary browser, so it just stands out to me that you went the other way :smile:

tests/features/00_smoke.feature
```diff
@@ -0,0 +1,6 @@
+Feature: Smoke test
+  Simple test to check if test site is running.
+
+  Scenario: See front page
+    Given I am on "/"
+    Then I should see "Biblo"
```
kallehauge, line 6: Is it on purpose that there's "only" a feature and no step definitions?

composer.json
kasperg, line 29: No - Drupal Core tests fail if its not included.

composer.json
kasperg, line 41: Linting catches syntax errors. PHPCS catches code style deviations.

This is primarily for the deployment pipeline where we could deploy something that does not follow our code style but we should not deploy something that does not compile.

docker-compose.yml
kasperg, line 77: The Selenium Chrome Docker image died on me right after booting.

tests/features/00_smoke.feature
kasperg, line 6: Yep. This feature was primarily meant to test that the Behat + Selenium setup is working.

PR #20, opened 2016-03-02
Add dummy data to the service DB

kallehauge:
Provide dummy data in the Service DB Container. The data can be expanded upon further down the road if needed. This gives us some data so we at least have something to pull out of the service when developing.

Data:
- 2x users (André & Kasper).
- 2x groups with members associated to it.
- 2x posts (one with a comment).
- 2x user roles (granted to the different users).
- 1x flag (André Flagged a comment from Kasper).
- 1x quarantine (Kasper got suspended for 1 year).

**_OBS!_** The postgres image also looks for files in the `docker-entrypoint-initdb.d` directory like MariaDB but it doesn't look for `*.gz` files, so we have to use a `*.sql` file instead, unless we make our own image that extracts the file when starting a container (but I don't find the extra work to be worth it).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -52,6 +52,8 @@ service_db:
   image: postgres
   ports:
     - '5432/tcp'
+  volumes:
+    - './docker/service_db/db.sql:/docker-entrypoint-initdb.d/db.sql:ro'
   environment:
     POSTGRES_DB: db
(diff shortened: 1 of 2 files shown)


kasperg:
:+1:

Using a SQL file is generally fine. @bicschneider has also recommended doing so.

PR #21, opened 2016-03-03
Use global path filter

kasperg:
The tool specific one does not seem to exclude our client library.


kallehauge:
@kasperg I guess we don't need the 'eslint` filter, when we generalise from the tool specific filtering?

kasperg:
@kallehauge Agreed. Removed.

PR #22, opened 2016-03-04
Page Manager & DBC Community Development module

kallehauge:
- Add and enable Page Manager module _(The `page_manager.page.node_view.yml` is a default page the module offers and normally enables. I have disabled it, since we don't use nodes anyway)_.
- Add DBC Community Service Development module.

_**Edit:** This PR has expanded a bit in size. Because the Page Manager module doens't make much sense right now, without the UI, we combine a development module and Page manger to one PR._

### DIFF
--- a/web/config/sync/core.extension.yml
+++ b/web/config/sync/core.extension.yml
@@ -2,6 +2,7 @@ module:
   basic_auth: 0
   block: 0
   breakpoint: 0
+  ctools: 0
   dbcdk_community: 0
   dblog: 0
   dynamic_page_cache: 0
@@ -11,6 +12,8 @@ module:
   link: 0
   node: 0
   page_cache: 0
+  page_manager: 0
+  page_manager_ui: 0
   rest: 0
   serialization: 0
   shortcut: 0
--- a/web/config/sync/core.extension.yml
+++ b/web/config/sync/core.extension.yml
@@ -13,7 +13,6 @@ module:
   node: 0
   page_cache: 0
   page_manager: 0
-  page_manager_ui: 0
   rest: 0
   serialization: 0
   shortcut: 0
--- a/web/config/sync/page_manager.page.node_view.yml
+++ b/web/config/sync/page_manager.page.node_view.yml
@@ -1,0 +1,17 @@
+uuid: 75787784-0219-4300-bd00-dc4e5e1fc7fc
+langcode: en
+status: false
+dependencies: {  }
+_core:
+  default_config_hash: 1rbdWeEeBjGARzFlodw9zcw3FlX6WQD83Rsckxl9_3o
+id: node_view
+label: 'Node view'
+use_admin_theme: false
+path: '/node/{node}'
+access_logic: and
+access_conditions: {  }
+parameters:
+  node:
+    machine_name: node
+    type: 'entity:node'
+    label: Node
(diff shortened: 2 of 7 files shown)


kallehauge:
The `page_manager.page.node_view.yml` is a default page the module offers and normally enables. I have disabled it, since we don't use nodes anyway.

web/config/sync/core.extension.yml
```diff
@@ -11,6 +12,8 @@ module:
   link: 0
   node: 0
   page_cache: 0
+  page_manager: 0
+  page_manager_ui: 0
```
kasperg: Should we have this enabled as a part of the default config?

I am thinking that it is only needed in development scenarios and we want to keep the default enabled modules to a minimum.

web/config/sync/page_manager.page.node_view.yml
```diff
@@ -0,0 +1,17 @@
+uuid: 75787784-0219-4300-bd00-dc4e5e1fc7fc
+langcode: en
+status: false
```
kasperg, line 3: :+1:

kasperg:
:+1: for this. A single question/comment.

web/config/sync/core.extension.yml
kallehauge: I get your point but unlike ex `Devel` and other development tools, this only presents a UI that will be hidden for non-administrator users (I've always had the same track of thought with Views UI). But it's a good question. I don't mind having to enable it during development.

kallehauge:
@kasperg I have added a "Development module" that will enable the Page Manager UI, if enabled.
We talked (face to face) about keeping the "Debug mode" out of the Community Service UI but due to the amount of information it spits out when a service-call is made (see picture below), this seems more like a _"should only be enabled when someone actually has to debug the API"_ and the development module is intended to be enabled at all time during development due to UI module dependencies etc.

screen shot 2016-03-06 at 15 45 16

Because of the reasons above then I simply expanded the admin form with an "Advanced Options" collapsible box since it's the cleanest method and if the user is able to set the _Host Url_ to begin with then the user has some sort of "administration role" and should know that it is something that should not be enabled in production.

_**Edit:** And the settings.php file is ":ro" when we use Docker. This prevents us from quickly enabling/disabling the configuration in the `settings.php` file and we would have to remember the name of the configuration or look it up in a development configuration service that replaces the original Client API Configuration service._

web/modules/custom/dbcdk_community/src/CommunityApiConfiguration.php
```diff
@@ -26,10 +26,19 @@ public function __construct() {
...
     // so we make sure the host does not end with one.
     $host = rtrim($config->get('community_service_url'), '/');
     $this->setHost($host);
+
+    // Set Debug Mode if it is enabled in the admin form.
+    // The form type "checkbox" return 0 or 1 as its values for true and false
+    // but the API says it requires a boolean, so we cannot simply use the
+    // value and therefor have to define it ourselves.
+    if ($config->get('community_service_debug')) {
+      $this->setDebug(TRUE);
```
kasperg: You could cast is to a boolean:

``` php
$this->setDebug((bool) $config->get('community_service_debug'));
```

Another option would be to cast it when setting the configuration value in the submit handler.

kasperg:
I agree about you comments. Approved and merged.

PR #24, opened 2016-03-07
Initial version of data generator

kasperg:
This will generate some dummy data with users, posts, comments, flags
and quarantines.

### DIFF
--- a/web/modules/custom/dbcdk_community_devel/src/Command/GenerateCommand.php
+++ b/web/modules/custom/dbcdk_community_devel/src/Command/GenerateCommand.php
@@ -1,0 +1,191 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\dbcdk_community_devel\Command\GenerateCommand.
+ */
+
+namespace Drupal\dbcdk_community_devel\Command;
+
+use DBCDK\CommunityServices\Api\CommentApi;
+use DBCDK\CommunityServices\Api\FlagApi;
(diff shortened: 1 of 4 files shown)


kasperg:
@kallehauge Would you review this please.

Scrutinizer fails because of duplicated code notices. I would be happy to discuss how we can improve this.

web/modules/custom/dbcdk_community_devel/src/Command/GenerateCommand.php
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function execute(InputInterface $input, OutputInterface $output) {
+    $io = new DrupalStyle($input, $output);
+
+    $configuration = \Drupal::service('dbcdk_community.api.configuration');
+    $api_client = new ApiClient($configuration);
```
kallehauge: Why is the configuration fetched as a service, but not the client? - The same question withstands for the different API's that is also defined as services? This would eliminate the need to call the ApiClient completely :relaxed:

kallehauge:
@kasperg I don't believe it's an error but instead an edge-case that is causing Scrutinizer to fail. The code looks fine and it's the `foreach(range(1, 10) as $i)` it complains about.
I guess we could work around it by upping the amount of posts being created to 20?

kallehauge:
:+1: You can merge when we fetch the API's as services

kasperg:
The APIs are fetched as services now. I'll merge.

PR #25, opened 2016-03-09
#486: Flagged content list

kasperg:
This PR adds a page listing content with unread flags.

PR #26, opened 2016-03-14
Feature/profiles

kallehauge:


web/config/sync/core.extension.yml
```diff
@@ -2,11 +2,14 @@ module:
...
   block: 0
   breakpoint: 0
+  ckeditor: 0
   ctools: 0
   dbcdk_community: 0
   dblog: 0
   dynamic_page_cache: 0
+  editor: 0
   field: 0
+  file: 0
```
kallehauge, line 12: CKEditor requires the file module because of the "Upload image" functionality.

kasperg, line 12: :+1:

web/config/sync/filter.format.dbcdk_community_profiles__description.yml
```diff
@@ -0,0 +1,23 @@
+uuid: bda9e83c-d51b-40f6-a306-6cd8b0496666
+langcode: en
+status: true
+dependencies: {  }
+name: 'DBCDK Community Profiles - Description'
+format: dbcdk_community_profiles__description
```
kasperg, line 6: Does it make sense to create a specific format for each field as indicated by this name? Could we provide just one for rich content from the service?

web/config/sync/page_manager.page.profile.yml
```diff
@@ -0,0 +1,15 @@
+uuid: c8349b6a-d910-491d-961a-2667be7aae70
+langcode: en
+status: true
+dependencies: {  }
+id: profile
```
kasperg: We talked about prefixing these page manager variant with the module names.

web/config/sync/page_manager.page.profiles.yml
```diff
@@ -0,0 +1,11 @@
+uuid: 162c3bb8-a66d-4bfa-ac0a-1012d65206cb
+langcode: en
+status: true
+dependencies: {  }
+id: profiles
```
kasperg: See above

web/config/sync/page_manager.page.profile.yml
```diff
@@ -0,0 +1,15 @@
...
+label: Profile
+use_admin_theme: true
+path: '/admin/content/dbcdk-community/profiles/{username}'
+access_logic: and
+access_conditions: {  }
```
kasperg: We should add permissions and provide access control based on these here.

web/config/sync/page_manager.page_variant.profile.yml
```diff
@@ -0,0 +1,33 @@
+uuid: 34531563-6643-435e-bd00-cc599ffc16b7
+langcode: en
+status: true
+dependencies:
+  config:
+    - page_manager.page.profile
+  module:
+    - dbcdk_community
+id: profile
```
kasperg: See above.

web/config/sync/page_manager.page_variant.profiles.yml
```diff
@@ -0,0 +1,33 @@
+uuid: 44f8933e-e7b9-4d8b-be00-d5954e05b116
+langcode: en
+status: true
+dependencies:
+  config:
+    - page_manager.page.profiles
+  module:
+    - dbcdk_community
+id: profiles
```
kasperg: See above

web/modules/custom/dbcdk_community/src/CommunityTraits.php
```diff
@@ -0,0 +1,39 @@
...
+   *
+   * @param \DBCDK\CommunityServices\ApiException $e
+   *   A caught Community Service API Exception.
+   */
+  protected function handleException(ApiException $e) {
+    // Set a more human-friendly error message for users.
+    drupal_set_message($this->t('An error occurred when the system tried to fetch data from an external service. Please try and refresh the page or contact an administrator.'), 'error');
+
+    // Log errors to dblog.
+    \Drupal::logger('DBCDK Community Service')->error(urldecode($e->getMessage()) . '<br><br>' . $e->getTraceAsString());
```
kasperg: According to the PHP documentation [casting the exception as a string should output just about this](http://php.net/manual/en/exception.tostring.php). Have you tried this?

I do not think having HTML-markup in the log message is not a good thing as it makes assumptions about how the logged content is rendered.

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -0,0 +1,269 @@
...
+   * Creates a Profile Edit Form instance.
+   *
+   * @param RequestStack $request_stack
+   *   The current request stack.
+   * @param ProfileApi $profile_api
+   *   The DBCDK Community Service Profile API.
+   */
+  public function __construct(RequestStack $request_stack, ProfileApi $profile_api) {
+    $this->profileApi = $profile_api;
+    $this->requestStack = $request_stack;
```
kasperg: The RequestStack only seems to be needed for object creation. How about changing the contructor signatur to `Profile $profile, ProfileApi $profile_api` and handling the rest in `create`?

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -0,0 +1,269 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state, $username = NULL) {
+    // Defensive coding to make sure the editor of the profile does not submit
+    // a form without default values (this could cause deletion of data if the
+    // editor was unaware of the lack of values).
+    if (empty($this->profile)) {
+      drupal_set_message($this->t('The system could not find any information about the profile at this moment. Please try and refresh or contact an administrator.'), 'error');
```
kasperg, line 100: Would this error message be generated twice if profile retrieval fails (see call to handleExceptio L264)?

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -0,0 +1,269 @@
...
+
+    if (!empty($new_data)) {
+      try {
+        // The Community Service needs to know who is being altered, so we have
+        // to set the profiles ID to the body of the PUT request.
+        $new_data['id'] = $this->profile->getId();
+        $this->profile = $this->profileApi->profileUpsert(json_encode($new_data));
+        drupal_set_message($this->t('The profile "%profile" have been updated.', ['%profile' => $this->profile->getUsername()]));
+      }
+      catch (\Exception $e) {
```
kasperg: Should this be `ApiException`?

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -0,0 +1,269 @@
...
+      }
+    }
+
+    if (!empty($new_data)) {
+      try {
+        // The Community Service needs to know who is being altered, so we have
+        // to set the profiles ID to the body of the PUT request.
+        $new_data['id'] = $this->profile->getId();
+        $this->profile = $this->profileApi->profileUpsert(json_encode($new_data));
+        drupal_set_message($this->t('The profile "%profile" have been updated.', ['%profile' => $this->profile->getUsername()]));
```
kasperg: Nitpicking typo: The profile _has_.

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -0,0 +1,269 @@
...
+      $filter = [
+        'limit' => 1,
+        'where' => [
+          'username' => $username,
+        ],
+      ];
+
+      // Since we have a limit of 1 result, we simply select that one result
+      // from the results array instead of looping through it.
+      return $this->profileApi->profileFind(json_encode($filter))[0];
```
kasperg: What happens if the result is empty? I would think we would get a warning trying to get an invalid entry or the like.

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -0,0 +1,269 @@
...
+   * @var Profile $profile
+   */
+  protected $profile;
+
+  /**
+   * The date format the Community Client expects.
+   *
+   * @var string $dateFormat
+   */
+  protected $dateFormat = 'Y-m-d';
```
kasperg: This should be a class constant instead of a variable.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
```diff
@@ -0,0 +1,247 @@
...
+    catch (\Exception $e) {
+      $this->handleException($e);
+    }
+
+    // Defensive coding to make sure we don't break anything if the API returns
+    // "NULL" or something similar instead of an array of profiles.
+    // Defensive coding to make sure we don't cause a fatal error by processing
+    // the results from the request as a profile, without it being a profile
+    // object - The results could return NULL.
+    if (!empty($profile) && $profile instanceof Profile) {
```
kasperg: If we want to code defensively then checking `$profile instanceof Profile` should be sufficient guard against `NULL` values.

web/config/sync/filter.format.dbcdk_community_profiles__description.yml
kallehauge, line 6: I wasn't sure about the different needs in each field, so this was just a more cautious decision - I would prefer to make one that is generic enough to be used in all fields that has markup but I think we would have to discuss this with the Biblo.dk team to be sure of any generalisations

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
```diff
@@ -0,0 +1,247 @@
...
+          break;
+
+        default:
+          // Use the machine-name of the field to fetch its value from the
+          // Profile object (format: $profile->getFieldName()).
+          $value = $profile->{'get' . ucfirst($field)}();
+
+          // The render array cannot take an object as a column value (this
+          // will cause a fatal error). Instead we check if it is an object
+          // and return an empty string and log the event.
```
kasperg: Can we cast it to be a string e.g. `$value = (string) $profile->{'get' . ucfirst($field)}();`?

web/config/sync/page_manager.page.profile.yml
kallehauge: Waiting for pull #25 to be merged due to patches to allows permissions in conditions

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
```diff
@@ -0,0 +1,247 @@
...
+          break;
+
+        // We have to allow HTML in the output of the description field since
+        // Twig automatically escapes string variables because markup is meant
+        // to be handled in templates.
+        case 'description':
+          $value = [
+            'data' => [
+              '#markup' => $profile->getDescription(),
+              '#allowed_tags' => Xss::getAdminTagList(),
```
kasperg, line 171: :+1:

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
```diff
@@ -0,0 +1,247 @@
...
+   * @param string $plugin_id
+   *   The plugin_id for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \DBCDK\CommunityServices\Api\ProfileApi $profile_api
+   *   The DBCDK Community Service Profile API.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, ProfileApi $profile_api) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->profileApi = $profile_api;
```
kasperg, line 55: Would it be possible to load the Profile in `create` method and just having a Profile as a variable?

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
```diff
@@ -0,0 +1,277 @@
...
+   */
+  public function build() {
+    // Tries to fetch profiles from the Community Service.
+    // We catch any API Exceptions and stop the rest of the code from executing
+    // so we don't get fatal errors the will cause a "status code 500" but
+    // rather displays an empty table.
+    try {
+      $filter = [
+        'limit' => $this->pagerLimit,
+        'offset' => $this->requestStack->getCurrentRequest()->query->get('page') * $this->pagerLimit,
```
kasperg: Calling `pager_default_initialize()` will return the page. We are bound by this method anyway later on so I think we might as well use it.

This also means we can remove the dependency to the requeststack.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
kallehauge, line 55: There is no context yet in `__construct` and it is not possible to access `$this` inside the static create method - so I see no other option for us but to load the Profile in the `build` method

web/modules/custom/dbcdk_community/src/CommunityTraits.php
kallehauge: You're correct. I tried to fetch more specific values by using `$e->getMessage()`, `$e->getTrace()`etc. when I should just cast the exception as it is. - This also ended up removing the need for the weird `handleFormat` method in the `CommunityTrait` and therefor removing the need for the trait completely.

And this also eliminated the HTML in the log message (btw: it was a good point with the assumption about how the log is being rendered :+1: )

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
kallehauge, line 100: The `handleException` method has been removed as stated in one of the comments above. But yes, that was previously the case - Amazing eyes! :eyes:

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
kallehauge: The `empty` check was also to make sure the variable was set because the profile variable was only set inside the the `try {}` - but this is no longer the case. Good catch!

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
kallehauge: It will cause a fatal error if the value happened to be i.e. an object since it cannot be converted to a string

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
kallehauge: If the `pager_default_initialize()` function gets refactored to "not being a global thingy" and we have to refactor the code later on (thinking Drupal 8.2.x, 8.3.x...), then there would be less work in changing it because we got the page-value the "correct" way to begin with?

kallehauge:
This PR is now handled in PR #27 instead.

PR #27, opened 2016-03-15
Feature/profiles-rebased

kallehauge:
PR Contains:
- Table of community profiles
- Single display of a community profile
- Edit form for a community profile

_The PR also includes a basic markup text format that is no loger needed by the profile description field - but I agreed with @kasperg that it made sense to generalise it and include it anyways (for later use)_


kasperg:
Great!

PR #28, opened 2016-03-15
Correct code standards for Community Content

kallehauge:
My PHP CodeSniffer with Drupal standards complained about:
- some declarations of `use` that... well.. was not being used :smile: 
- Missing comma's at the end of arrays.
- Punctuation in DocBlock titles.
- And the whitespace left and right of single value array elements that we discussed earlier @kasperg.

### DIFF
--- a/web/modules/custom/dbcdk_community/src/Plugin/Block/FlaggedContentList.php
+++ b/web/modules/custom/dbcdk_community/src/Plugin/Block/FlaggedContentList.php
@@ -9,9 +9,7 @@ namespace Drupal\dbcdk_community\Plugin\Block;
 
 use DBCDK\CommunityServices\ApiException;
 use Drupal\Core\Block\BlockBase;
-use Drupal\Core\Logger\LoggerChannelFactory;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
-use Drupal\dbcdk_community\Content\FlaggableContent;
 use Drupal\dbcdk_community\Content\FlaggableContentRepository;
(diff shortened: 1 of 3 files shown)


web/modules/custom/dbcdk_community/src/Plugin/Block/FlaggedContentList.php
```diff
@@ -9,9 +9,7 @@
 
 use DBCDK\CommunityServices\ApiException;
 use Drupal\Core\Block\BlockBase;
-use Drupal\Core\Logger\LoggerChannelFactory;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
-use Drupal\dbcdk_community\Content\FlaggableContent;
```
kasperg, line 14: These are used in DocBlocks.

If we delete them here then we need to use fully qualified classnames there.

kallehauge, line 14: Hmm, I think we have to go with the fully qualified names then (not too keen about it though). But the trade off is a PHP CodeSniffer that will complain about unused classes

kasperg:
@kallehauge: Thanks for going through this.

The primary question is whether we should use fully qualified classnames in DocBlocks and remove them from use statements or the other way around.

kallehauge:
Not a problem!

The main "problem" is Codesniffer - In the code I've written, I've used the classname with using "use" and fully qualified names - So I might have to refactor some DocBlocks when we decide what to do :smile:

kasperg:
Let's refactor the DocBlocks then.

PR #29, opened 2016-03-15
Use local coding standards

kasperg:
Coding standards may differ between environments (different versions of Codesniffer, different versions of Coding Standards).

This PR includes a shared version of both and uses these for CI purposes.


kasperg:
This should at least be workable and usable locally. What do yo think @kallehauge?

kallehauge:
I like it! - The code standards corrections for the generate command is a bit out of scope from the ticket, but that doesn't matter.

@kasperg you can merge this when you wish if you agree on not adding any CI changes for now due to technical complications

PR #30, opened 2016-03-16
Set xdebug.max_nesting_level to meet install.php requirements

kasperg:
If this is not set we are not able to complete the Drupal installer.


kallehauge:
Brilliant :+1:

PR #31, opened 2016-03-16
Links from flagged content to community site

kasperg:


kasperg:
This is ready for review now @kallehauge

web/modules/custom/dbcdk_community/src/Form/SettingsForm.php
```diff
@@ -44,14 +44,37 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
-      ),
+    $form['community_service_url'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('Community Service Url'),
+      '#default_value' => $config->get('community_service_url'),
+      '#description' => $this->t('The Community Service Url until the endpoints start. Example: %example', array('%example' => 'https://dbcdk-community.dk/api/')),
+    ];
+
+    // Vertical tab for community site settings.
+    $form['community_site'] = array(
```
kallehauge: If you find any reason to "correct" this, feel free to do so, but I really dont mind. But I can see that you've changed "community service url" and "community service debug" to use `[]` instead of `array()` when you removed the unnecessary nesting.

In short: you're using `array()` here. Change it if you feel like it.

kasperg: This is fixed now.

kallehauge:
Looks good :+1:

PR #32, opened 2016-03-17
Exposed filter form for Community Profiles list

kallehauge:
Exposed filter form (it, currently, "only" allows to search for a profile by username)
- Add new ProfileFilterForm to be used in ProfilesBlock::build().
- Build and expose the ProfileFilterForm in ProfilesBlock::build().
- Allow null values in ProfilesBlock::buildTable(). _Reason: If we try to filter the service query with a where condition and the request yields no results then the community service returns NULL. - Because of this we have to support NULL values with a successful request from the community service._

### DIFF
--- a/web/modules/custom/dbcdk_community/src/Form/ProfilesFilterForm.php
+++ b/web/modules/custom/dbcdk_community/src/Form/ProfilesFilterForm.php
@@ -1,0 +1,61 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\dbcdk_community\Form\ProfilesFilterForm.
+ */
+
+namespace Drupal\dbcdk_community\Form;
+
+use Drupal\Core\Form\FormBase;
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Exposed filter form for the Community Profiles listing page.
+ */
+class ProfilesFilterForm extends FormBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'dbcdk_community_profiles_filter_form';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $search_query = $this->getRequest()->get('search');
+
+    $form['filter'] = [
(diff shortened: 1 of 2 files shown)


web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
```diff
@@ -121,6 +144,13 @@ public function build() {
...
         'offset' => $this->pageNumber * $this->pagerLimit,
       ];
+      if (!empty($this->searchQuery)) {
+        $filter['where'] = [
+          'username' => [
+            'like' => '%' . $this->searchQuery . '%',
+          ],
+        ];
+      }
       $profiles = $this->profileApi->profileFind(json_encode($filter));
```
kasperg: The client will return `NULL` for empty results. This can lead to unelegant type handling later on.

The smartest way to handle this at the moment is to cast the results to be arrays. `NULL` results will be converted to an empty array.

web/modules/custom/dbcdk_community/src/Form/ProfilesFilterForm.php
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $search_query = $this->getRequest()->get('search');
+
+    $form['filter'] = [
+      '#type' => 'details',
+      '#title' => $this->t('Filter list'),
+      '#open' => empty($search_query) ? FALSE : TRUE,
```
kasperg, line 34: Nice detail :+1:

kasperg:
@kallehauge: Review complete. Looks great!

I have a suggestion about how to fix the `NULL` values from the library. Otherwise it's good to go!

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
kallehauge: Brilliant idea - I'll fix this later :relaxed:

PR #33, opened 2016-03-22
Profile roles

kallehauge:
- Correct API Class name for Community Roles API Service
- Add visual presentation of a profile's roles on the single profile display page.
- Small refactoring of the submitForm() method to make the method work more like a skinny controller.
- Add functionality to add and remove roles from profiles.

### DIFF
--- a/web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
+++ b/web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
@@ -9,6 +9,7 @@ namespace Drupal\dbcdk_community\Form;
 
 use DBCDK\CommunityServices\ApiException;
 use DBCDK\CommunityServices\Api\ProfileApi;
+use DBCDK\CommunityServices\Model\Profile;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
 use Drupal\Core\Form\FormBase;
@@ -28,57 +29,76 @@ class ProfileEditForm extends FormBase implements ContainerInjectionInterface {
   /**
    * The DBCDK Community Service Profile API.
    *
-   * @var ProfileApi $profileApi
+   * @var \DBCDK\CommunityServices\Api\ProfileApi $profileApi
    */
   protected $profileApi;
 
   /**
    * The Community Service Profile object from the username context.
    *
-   * @var Profile $profile
+   * @var \DBCDK\CommunityServices\Model\Profile $profile
    */
   protected $profile;
 
   /**
+   * Array of all defined community roles.
+   *
+   * @var \DBCDK\CommunityServices\Model\CommunityRole[] $communityRoles
+   */
+  protected $communityRoles;
+
+  /**
    * Creates a Profile Edit Form instance.
    *
-   * @param null|Profile $profile
-   *   The current Community Profile or NULL if we had no successful matches.
-   * @param ProfileApi $profile_api
+   * @param \DBCDK\CommunityServices\Model\Profile $profile
+   *   The DBCDK Community Profile object we wish to edit.
+   * @param \DBCDK\CommunityServices\Model\CommunityRole[] $community_roles
+   *   Results array with all possible Community Roles.
+   * @param \DBCDK\CommunityServices\Api\ProfileApi $profile_api
    *   The DBCDK Community Service Profile API.
    */
-  public function __construct($profile, ProfileApi $profile_api) {
+  public function __construct(Profile $profile, array $community_roles, ProfileApi $profile_api) {
     $this->profileApi = $profile_api;
-    // Set the profile as NULL if we did not get any results array from the
-    // service - or get the first element of the array, since we only get one
-    // result, and set this as our profile property.
-    $this->profile = empty($profile) ? NULL : $profile[0];
+    $this->communityRoles = $community_roles;
+    $this->profile = $profile;
   }
 
   /**
    * {@inheritdoc}
    */
   public static function create(ContainerInterface $container) {
-    // Get the Profile API and fetch the Profile we wish to edit and inject
-    // this into the forms constructor.
     $profile_api = $container->get('dbcdk_community.api.profile');
+    $community_roles_api = $container->get('dbcdk_community.api.community_roles');
+
     try {
+      // Get the profile we wish to alter with the form.
       $filter = [
-        'limit' => 1,
         'where' => [
           'username' => $container->get('request_stack')->getCurrentRequest()->get('username'),
         ],
       ];
-      /* @var Profile $profile */
-      $profile = $profile_api->profileFind(json_encode($filter));
+      /* @var \DBCDK\CommunityServices\Model\Profile[] $profile */
+      $profile = $profile_api->profileFindOne(json_encode($filter));
+      // The Swagger-compiled Community Profile Model does not support Community
+      // Roles when using "include" to get the profile. The response from the
+      // service includes it but the model does not care about it, so we have to
+      // handle this with a separate call.
+      $profile->communityRoles = (array) $profile_api->profilePrototypeGetCommunityRoles($profile->getId());
+
+      // Get all possible Community Roles. This will be used to give a moderator
+      // the possibility to grant roles the profile does not yet have.
+      /* @var \DBCDK\CommunityServices\Model\CommunityRole[] $community_roles */
+      $community_roles = (array) $community_roles_api->communityRoleFind();
     }
     catch (ApiException $e) {
       \Drupal::logger('DBCDK Community Service')->error($e);
-      $profile = NULL;
+      $profile = [];
+      $community_roles = [];
     }
 
     return new static(
       $profile,
+      $community_roles,
       $profile_api
     );
   }
@@ -139,11 +159,23 @@ class ProfileEditForm extends FormBase implements ContainerInjectionInterface {
       '#date_date_format' => self::DATE_FORMAT,
     ];
 
+    $form['roles'] = [
+      '#type' => 'checkboxes',
+      '#title' => $this->t('Roles'),
+      '#default_value' => array_map(function($role) {
+        return $role->getId();
+      }, $this->profile->communityRoles),
+      '#options' => array_reduce($this->communityRoles, function($result, $role) {
+        $result[$role->getId()] = $role->getName();
+        return $result;
+      }),
+    ];
+
     $form['description'] = [
       '#type' => 'text_format',
       '#title' => $this->t('Description'),
       '#default_value' => $this->profile->getDescription(),
-      '#format' => 'dbcdk_community_profiles__description',
+      '#format' => 'plain_text',
     ];
 
     $form['actions']['#type'] = 'actions';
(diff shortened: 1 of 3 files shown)


kallehauge:
Scrutinizer complains about indentation but I am certain that it is correct and I can run PHPCS with Drupal standards with it complaining.

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -28,57 +28,79 @@ class ProfileEditForm extends FormBase implements ContainerInjectionInterface {
...
       $filter = [
         'limit' => 1,
         'where' => [
           'username' => $container->get('request_stack')->getCurrentRequest()->get('username'),
         ],
       ];
-      /* @var Profile $profile */
-      $profile = $profile_api->profileFind(json_encode($filter));
+      /* @var \DBCDK\CommunityServices\Model\Profile[] $profile */
+      $profile = (array) $profile_api->profileFind(json_encode($filter));
```
kasperg: How about using [`profileFindOne`](https://github.com/DBCDK/biblo-admin/blob/develop/lib/client/src/Api/ProfileApi.php#L796) instead? This will only return one `Profile` instance.

If not then we should reduce the return value of `profileFind` to a single object as soon as possible to avoid checking and passing around arrays later on.

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -28,57 +28,79 @@ class ProfileEditForm extends FormBase implements ContainerInjectionInterface {
...
+      // The Swagger-compiled Community Profile Model does not support Community
+      // Roles when using "include" to get the profile. The response from the
+      // service includes it but the model does not care about it, so we have to
+      // handle this with a separate call.
+      if (!empty($profile)) {
+        $profile[0]->communityRoles = (array) $profile_api->profilePrototypeGetCommunityRoles($profile[0]->getId());
+      }
```
kasperg: Worth noting here is also that the `Profile` class does not have a communityRoles property or accessors so we just stick on a new public property.

I think that is worth fixing e.g. by implementing a decorator for the generated Profile class which also supports adding roles or we might be able to do something cool with traits but that can wait for now.

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -139,6 +161,18 @@ public function buildForm(array $form, FormStateInterface $form_state, $username
...
     ];
 
+    $form['roles'] = [
+      '#type' => 'checkboxes',
+      '#title' => $this->t('Roles'),
+      '#default_value' => array_map(function($role) {
+        return (int) $role->getId();
+      }, $this->profile->communityRoles),
+      '#options' => array_reduce($this->communityRoles, function($result, $role) {
+        $result[(int) $role->getId()] = (string) $role->getName();
```
kasperg: Any reason for the cast here? Are the types declared in the object not enougth?

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -139,6 +161,18 @@ public function buildForm(array $form, FormStateInterface $form_state, $username
...
+        return $result;
```
kasperg: Using `array_reduce` to transform an keyed array to another was a bit surprising to me. I always saw the return value of this as a primitive (product of all values, all strings joined etc.) but I guess this works as well :+1: .

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -222,23 +297,61 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
-    ]);
+  /**
+   * Update a Community Profile's roles.
+   *
+   * Find out if we have to remove or add any roles from a Community Service
+   * based on values from $form_state and the current Community Profiles roles
+   * and return a boolean representing if anything were updated or not.
+   *
+   * @param array $roles
+   *   The value of the roles input from form_state.
```
kasperg: So this is in fact role ids? If that is the case then perhaps update the variable name and comment to indicate that.

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -222,23 +297,61 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
+   *
+   * @return bool
+   *   Returns TRUE if there were any changes to update and vise versa.
+   */
+  protected function updateProfileRoles(array $roles) {
+    // Format an array of the profiles role ids before any update has happened.
+    $old_roles = array_map(function($role) {
+      return (int) $role->getId();
```
kasperg: Why the cast here?

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -222,23 +297,61 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
+   *
+   * @param array $roles
+   *   The value of the roles input from form_state.
+   *
+   * @return bool
+   *   Returns TRUE if there were any changes to update and vise versa.
+   */
+  protected function updateProfileRoles(array $roles) {
+    // Format an array of the profiles role ids before any update has happened.
+    $old_roles = array_map(function($role) {
```
kasperg: Consider renaming this to `$old_role_ids` for clarity.

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -222,23 +297,61 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
+      return (int) $role->getId();
+    }, $this->profile->communityRoles);
+
+    // Format an array of the role ids we want the profile to have.
+    $new_roles = array_filter($roles);
```
kasperg: And here `$new_role_ids`.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
```diff
@@ -206,6 +222,36 @@ protected function parseProfile(Profile $profile, array $fields) {
   }
 
   /**
+   * Get a Community Profile's Roles.
+   *
+   * @param \DBCDK\CommunityServices\Model\Profile $profile
+   *   The Community Profile.
+   *
+   * @return array $list_items
+   *   Return an array of the Profile's Community Roles.
```
kasperg: We do not specify the return value variable name

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -176,43 +212,82 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
     ];
+    $profile_fields = $this->updateProfileFields($fields, $form_state);
 
-    // Go through each field and check if there is any differences from the
-    // original value and the new value. If so, we add it to an array of new
-    // data. - We do this to make sure we actually have to PUT any data to the
-    // Community Service and provide a different message to the editor when
-    // nothing was updated or an actual update happened.
+    // Update roles based on the submitted "roles" input.
+    $profile_roles = $this->updateProfileRoles($form_state->getValue('roles'));
```
kasperg: Updating profile fields and roles may cause exceptions to be thrown. I think here is the right place to catch them and show an error message to the user.

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -222,23 +297,61 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
         // to set the profiles ID to the body of the PUT request.
         $new_data['id'] = $this->profile->getId();
         $this->profile = $this->profileApi->profileUpsert(json_encode($new_data));
```
kasperg: This may cause an `ApiException` to be thrown. I do not think there is much use catching it here. In that case we should add a `@throws ApiException` to the DocBlock.

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -222,23 +297,61 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
+      foreach ($add_roles as $role) {
+        $this->profileApi->profilePrototypeLinkCommunityRoles((int) $role, $this->profile->getId());
+      }
+    };
+
+    // We find the difference from the old roles and the new once to see if any
+    // roles have to be removed from the community profile.
+    if ($remove_roles = array_diff($old_roles, $new_roles)) {
+      foreach ($remove_roles as $role) {
+        $this->profileApi->profilePrototypeUnlinkCommunityRoles((int) $role, $this->profile->getId());
```
kasperg: This may cause an ApiException to be thrown. I do not think there is much use catching it here. In that case we should add a @throws ApiException to the DocBlock.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
```diff
@@ -173,6 +174,21 @@ protected function parseProfile(Profile $profile, array $fields) {
           ];
           break;
 
+        // The profile doesn't contain its roles so we cannot get the value as
+        // a regular property so we have to make a separate request for roles.
+        // The display of this field is also different than a single string
+        // value since we would like the roles to be presented as a list to make
+        // it easier overview.
```
kasperg, line 180: Appreciate the level of commenting here. Could the value be as in description to avoid the `render()` call?

``` php
$value = [
    'data' => [
        '#theme' => 'item_list',
        '#list_type' => 'ul',
        '#empty' => $this->t('This user has not roles.'),
        '#items' => $this->getProfileRoles($profile),
    ]
]
```

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
```diff
@@ -206,6 +222,36 @@ protected function parseProfile(Profile $profile, array $fields) {
...
   /**
+   * Get a Community Profile's Roles.
+   *
+   * @param \DBCDK\CommunityServices\Model\Profile $profile
+   *   The Community Profile.
+   *
+   * @return array $list_items
+   *   Return an array of the Profile's Community Roles.
+   */
+  protected function getProfileRoles(Profile $profile) {
```
kasperg: I am considering where it would make sense to move loading of roles up with the profile load to keep the data load together. However I think this can wait until we get a better approach to handling of profile roles.

kasperg:
@kallehauge: Review complete. Great job.

I think we should discuss how we can imple handling of roles at some point.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
kallehauge, line 180: This was a great catch! :+1:

I have implemented the change and this should now be considered an outdated comment :)

kasperg, line 180: :+1:

PR #34, opened 2016-03-22
Add requirements hook for service connection status

kasperg:
When configuration is stored in settings.php it can be hard to see what
config is actually used. This helps a bit.

status report biblo admin 2016-03-22 15-14-51


kallehauge:
Perfect! :+1: Should we link to the requirements page from the Community Settings page?

kasperg:
If we were to link from two pages I think we should link from the requirements page to the community settings page.

If we were to do something similar then we could introduce a validation step for the settings form which performed similar webservice call.

PR #35, opened 2016-03-22
Add link from profiles overview to the front end

kallehauge:
Add a link from the Community Profiles overview to the front end.

### DIFF
--- a/web/config/sync/dbcdk_community.settings.yml
+++ b/web/config/sync/dbcdk_community.settings.yml
@@ -1,3 +1,4 @@
 community_service_url: ''
 community_site_post_url_pattern: '/posts/[id]'
 community_site_comment_url_pattern: '/comments/[id]'
+community_site_profile_url_pattern: '/profil/[id]'
--- a/web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
+++ b/web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
@@ -16,6 +16,10 @@ use Drupal\Core\Block\BlockBase;
 use Drupal\Core\Form\FormBuilder;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Link;
+use Drupal\Core\Url;
+use Drupal\dbcdk_community\Url\ModelUrlGenerator;
+use Drupal\dbcdk_community\Url\PropertyUrlGenerator;
+use Drupal\dbcdk_community\Url\UrlGeneratorInterface;
 
 /**
  * Provides a 'ProfilesBlock' block.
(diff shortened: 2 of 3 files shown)


web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
```diff
@@ -253,6 +280,13 @@ protected function parseProfile(Profile $profile, array $columns) {
           ]);
           break;
 
+        // Generate community profile link based on a pattern defined on the
+        // settings page.
+        case 'community_link':
+          $community_site_url = Url::fromUri($this->urlGenerator->generate($profile));
+          $row[] = Link::fromTextAndUrl($this->t('Biblo.dk'), $community_site_url);
```
kallehauge: @kasperg I use "Biblo.dk" as my text for the link to the front end here. Should we agree on some sort of standard? I can see we use `See on community site` on Flagged Content

kasperg: We should agree on a standard. You decide which you prefer.

About url/link generation is calling these static methods the way to go? [It seems like the way to go compared to the complexity of using an injected LinkGenerator](https://github.com/DBCDK/biblo-admin/blob/develop/web/modules/custom/dbcdk_community/src/Plugin/Block/FlaggedContentList.php#L172-L175). These classes do not seem to hook into other parts of the Drupal architecture anyway.

kallehauge: After verbal communication: We're changing to "View on Biblo.dk".

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
```diff
@@ -253,6 +280,13 @@ protected function parseProfile(Profile $profile, array $columns) {
           ]);
           break;
 
+        // Generate community profile link based on a pattern defined on the
+        // settings page.
+        case 'community_link':
+          $community_site_url = Url::fromUri($this->urlGenerator->generate($profile));
+          $row[] = Link::fromTextAndUrl($this->t('View on Biblo.dk'), $community_site_url);
```
kallehauge, line 287: @kasperg I have changed the text as we discussed :)

kasperg:
Great. If we also export the config for the profile pattern we should be good.

web/config/sync/dbcdk_community.settings.yml
```diff
@@ -1,3 +1,4 @@
 community_service_url: ''
 community_site_post_url_pattern: '/posts/[id]'
 community_site_comment_url_pattern: '/comments/[id]'
+community_site_profile_url_pattern: '/profile/[id]'
```
kasperg: Dette skal så også være `/profil/[id]`?

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
kasperg, line 287: :+1:

kasperg:
Merged.

PR #37, opened 2016-03-23
Fix profile description

kallehauge:
The description cannot contain any HTML at this moment so we do not wish to display any text formats.


kasperg:
:+1:

PR #38, opened 2016-03-29
Quarantines

kallehauge:
This PR contains the following changes:
- Add a Create Quarantine Form
- Add a Edit Quarantine Form
- Add a Quarantines Block to the Profile page
- Generalise DBCDK Community date format since the usage outgrew "birthdays"

### DIFF
--- a/web/config/sync/core.date_format.dbcdk_community_service_date.yml
+++ b/web/config/sync/core.date_format.dbcdk_community_service_date.yml
@@ -1,0 +1,8 @@
+uuid: 85e1d2d9-09e5-41de-8900-a8a9c4486a8b
+langcode: en
+status: true
+dependencies: {  }
+id: dbcdk_community_service_date
+label: 'DBCDK Community Service date'
+locked: false
+pattern: 'j. F Y'
--- a/web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
+++ b/web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
@@ -1,0 +1,228 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\dbcdk_community\Plugin\Block\ProfileQuarantinesBlock.
+ */
+
+namespace Drupal\dbcdk_community\Plugin\Block;
+
+use DBCDK\CommunityServices\Api\ProfileApi;
+use DBCDK\CommunityServices\Api\QuarantineApi;
+use DBCDK\CommunityServices\ApiException;
+use DBCDK\CommunityServices\Model\Quarantine;
+use Drupal\Component\Utility\Xss;
+use Drupal\Core\Datetime\DateFormatter;
+use Drupal\Core\Url;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Link;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a 'ProfileQuarantinesBlock' block.
+ *
+ * This block provides an overview of a Community Profile's Quarantines.
+ *
+ * @Block(
+ *   id = "dbcdk_community_profile_quarantines_block",
+ *   admin_label = @Translation("DBCDK Community Profile Quarantines"),
+ *   context = {
+ *     "username" = @ContextDefinition("string")
+ *   }
+ * )
+ */
+class ProfileQuarantinesBlock extends BlockBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * The DBCDK Community Service Profile API.
+   *
+   * @var ProfileApi $profileApi
+   */
+  protected $profileApi;
+
+  /**
+   * The DBCDK Community Service Quarantine API.
+   *
+   * @var QuarantineApi $quarantineApi
+   */
+  protected $quarantineApi;
+
+  /**
+   * Drupal's date formatter to format dates to Drupal Date Formats.
+   *
+   * @var \Drupal\Core\Datetime\DateFormatter $dateFormatter
+   */
+  protected $dateFormatter;
+
+  /**
+   * Creates a Profiles Block instance.
+   *
+   * @param array $configuration
+   *   A configuration array containing information about the plugin instance.
+   * @param string $plugin_id
+   *   The plugin_id for the plugin instance.
+   * @param mixed $plugin_definition
+   *   The plugin implementation definition.
+   * @param \DBCDK\CommunityServices\Api\ProfileApi $profile_api
+   *   The DBCDK Community Service Profile API.
+   * @param \DBCDK\CommunityServices\Api\QuarantineApi $quarantine_api
+   *   The DBCDK Community Service Quarantine API.
+   * @param \Drupal\Core\Datetime\DateFormatter $date_formatter
+   *   Drupal's date formatter to format dates to Drupal Date Formats.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, ProfileApi $profile_api, QuarantineApi $quarantine_api, DateFormatter $date_formatter) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->profileApi = $profile_api;
+    $this->quarantineApi = $quarantine_api;
+    $this->dateFormatter = $date_formatter;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('dbcdk_community.api.profile'),
+      $container->get('dbcdk_community.api.quarantine'),
+      $container->get('date.formatter')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    $context_username = $this->getContextValue('username');
+    $quarantines = [];
+    try {
+      // Fetch a Community Profile based on the username so we can the
+      // corresponding ID to fetch quarantines.
+      $profile_filter = [
+        'where' => [
+          'username' => $context_username,
+        ],
+      ];
+      $profile = $this->profileApi->profileFindOne(json_encode($profile_filter));
+
+      // Fetch quarantines.
+      $quarantine_filter = [
+        'order' => 'end DESC',
+        'where' => [
+          'quarantinedId' => $profile->getId(),
+        ],
+      ];
+      $quarantines = (array) $this->quarantineApi->quarantineFind(json_encode($quarantine_filter));
+    }
+    catch (ApiException $e) {
+      \Drupal::logger('DBCDK Community Service')->error($e);
+      $profile = NULL;
+    }
+
+    // Create an array of the fields we wish to display as columns in our table.
+    // The order the fields appear in the array is also the order they will
+    // be displayed.
+    $rows = [];
+    $fields = [
+      'reason' => $this->t('Reason'),
+      'start' => $this->t('Start date'),
+      'end' => $this->t('End date'),
+      'edit_link' => $this->t('Edit'),
+    ];
+    foreach ($quarantines as $quarantine) {
+      $rows[] = $this->parseQuarantine($quarantine, $fields, $context_username);
+    }
+
+    // Build table of quarantines.
+    $build = [];
+    $build['table'] = [
+      '#theme' => 'table',
+      '#header' => $fields,
+      '#rows' => $rows,
+      '#empty' => $this->t('There are no quarantines attached to this profile.'),
+      '#cache' => [
+        'max-age' => 0,
+      ],
+    ];
+
+    // Create quarantine link.
+    if (!empty($profile)) {
+      $build['add_quarantine'] = [
+        '#type' => 'link',
+        '#title' => $this->t('Create quarantine'),
+        '#url' => new Url('dbcdk_community.profile.quarantine.add', [
+          'username' => $profile->getUsername(),
+        ]),
+        '#attributes' => [
+          'class' => [
+            'button',
+            'button--danger',
+          ],
+        ],
+      ];
+    }
+
+    return $build;
+  }
+
+  /**
+   * Parse a Community Service Quarantine.
+   *
+   * Takes a Community Service Quarantine and an array of fields to determine
+   * what values that should be display in the table of quarantines.
+   * We also require a username so we can fill out the route arguments for the
+   * edit link.
+   *
+   * @param \DBCDK\CommunityServices\Model\Quarantine $quarantine
+   *   A Community Service Quarantine.
+   * @param array $fields
+   *   An array of fields in the following format: field_name => title.
+   * @param string $username
+   *   The username of the Community Profile.
+   *
+   * @return array $rows
+   *   An array of rows containing subjects and information.
+   */
+  protected function parseQuarantine(Quarantine $quarantine, array $fields, $username) {
+    $row = [];
+    foreach ($fields as $field => $title) {
+      switch ($field) {
+        // Format DateTime objects to a string.
+        case 'start':
+        case 'end':
+          $method = 'get' . ucfirst($field);
+          if (method_exists($quarantine, $method)) {
+            $row[] = $this->dateFormatter->format($quarantine->{$method}()->getTimestamp(), 'dbcdk_community_service_date');
+          }
+          break;
+
+        // We have to allow HTML in the output of the description field since
+        // Twig automatically escapes string variables because markup is meant
+        // to be handled in templates.
+        case 'reason':
+          $row[] = [
+            'data' => [
+              '#markup' => $quarantine->getReason(),
+              '#allowed_tags' => Xss::getAdminTagList(),
+            ],
+          ];
+          break;
+
+        // The edit_link field is not a field provided by the Community Service
+        // but a column we wish to display with a link to edit a quarantine.
+        case 'edit_link':
+          $row[] = Link::createFromRoute($title, 'dbcdk_community.profile.quarantine.edit', [
+            'username' => $username,
+            'quarantine_id' => $quarantine->getId(),
+          ]);
+          break;
+      }
+    }
+
+    return $row;
+  }
+
+}
--- a/web/config/sync/core.date_format.dbcdk_community_service_birthday.yml
+++ b/web/config/sync/core.date_format.dbcdk_community_service_birthday.yml
@@ -1,8 +1,0 @@
-uuid: f42c8c2f-fcfc-4639-8c1c-28e0cc39ccfe
-langcode: en
-status: true
-dependencies: {  }
-id: dbcdk_community_service_birthday
-label: 'DBCDK Community Service - Birthday'
-locked: false
-pattern: 'j. F Y'
--- a/web/config/sync/page_manager.page_variant.dbcdk_community_profile.yml
+++ b/web/config/sync/page_manager.page_variant.dbcdk_community_profile.yml
@@ -17,8 +17,18 @@ variant_settings:
       provider: dbcdk_community
       label_display: visible
       region: top
-      weight: 0
+      weight: -10
       uuid: 4a4a416a-f25b-463a-9228-406b075d5a7d
+      context_mapping:
+        username: username
+    206a0096-7fe9-47ff-aa18-24ac36511de7:
+      id: dbcdk_community_profile_quarantines_block
+      label: Quarantines
+      provider: dbcdk_community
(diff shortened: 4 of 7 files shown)


web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Create a Community Quarantine.
+ */
+class QuarantineCreateForm extends FormBase implements ContainerInjectionInterface {
+
+  /**
+   * The date format the Community Client expects.
+   */
+  const DATE_FORMAT = 'Y-m-d';
```
kasperg: This is not only the format the client expects but also the one we present to the user. I do not think it is a good idea to mix up the two.
- For the format we present to the user how about using `dbcdk_community_service_date`?
- For the service see my comment regarding the submit handler

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
```diff
@@ -0,0 +1,177 @@
...
+      $profile_api = $container->get('dbcdk_community.api.profile');
+      $profile = $profile_api->profileFindOne(json_encode($profile_filter));
+    }
+    catch (ApiException $e) {
+      \Drupal::logger('DBCDK Community Service')->error($e);
+      // Create an empty Profile instance. The constructor expects a Profile
+      // object and will cause a fatal error if it receives something else and
+      // we would like to display an empty page with a Drupal message in case of
+      // an error.
+      $profile = new Profile();
```
kasperg: Instead of creating an empty Profile we could set `$profile = NULL` and set the default value of `$profile` in the constructor to be NULL as well.

This is supported by PHP and we avoid odd checks for validity based on e.g. username. If we are to check for anything to determine if this is an existing profile I think it should be the id.

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
```diff
@@ -0,0 +1,177 @@
...
+    // form if we could find a Community Profile.
+    if (empty($this->profile->getUsername())) {
+      drupal_set_message($this->t('The system could not find the profile that should receive this quarantine. Please try again later or contact an administrator.'), 'error');
+      return $form;
+    }
+
+    $form['reason'] = [
+      '#type' => 'textarea',
+      '#title' => $this->t('Reason'),
+      '#required' => TRUE,
```
kasperg: This should be a textarea, right?

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
```diff
@@ -0,0 +1,177 @@
...
+      '#default_value' => date(self::DATE_FORMAT),
+      '#required' => TRUE,
+    ];
+
+    $form['end'] = [
+      '#type' => 'date',
+      '#title' => $this->t('End date'),
+      '#date_date_format' => self::DATE_FORMAT,
+      '#default_value' => date(self::DATE_FORMAT),
+      '#required' => TRUE,
```
kasperg: Is setting an end date required?

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
```diff
@@ -0,0 +1,177 @@
...
+      $quarantine = [];
+      foreach ($fields as $field) {
+        $quarantine[$field] = $form_state->getValue($field);
+      }
+
+      // Set the ID of the Drupal Moderator who created this quarantine.
+      $quarantine['creatorId'] = $this->currentUser()->id();
+      // Set the Community Profile ID that should own this quarantine.
+      $quarantine['quarantinedId'] = $this->profile->getId();
+      $this->quarantineApi->quarantineCreate(json_encode($quarantine));
```
kasperg: How about just creating a `Quarantine` object, setting values on that and passing that to `quarantineCreate`?

That way we get type completion for our accessors, type hinting for the values, handling of complex values e.g. DateTimes etc.

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
```diff
@@ -0,0 +1,177 @@
...
+
+      // Redirect to the Community Profile after successful submission.
+      $form_state->setRedirect('page_manager.page_view_dbcdk_community_profile', [
+        'username' => $this->profile->getUsername(),
+      ]);
```
kasperg: How about setting a message here?

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
```diff
@@ -0,0 +1,177 @@
...
+        'start',
+        'end',
+      ];
+      $quarantine = [];
+      foreach ($fields as $field) {
+        $quarantine[$field] = $form_state->getValue($field);
+      }
+
+      // Set the ID of the Drupal Moderator who created this quarantine.
+      $quarantine['creatorId'] = $this->currentUser()->id();
```
kasperg: I think it would be good to have a little more elaborate description of why we set information about the Drupal user here and not a Community service id.

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
```diff
@@ -0,0 +1,177 @@
...
+        'where' => [
+          'username' => $container->get('request_stack')->getCurrentRequest()->get('username'),
+        ],
+      ];
+      /* @var \DBCDK\CommunityServices\Api\ProfileApi */
+      $profile_api = $container->get('dbcdk_community.api.profile');
+      $profile = $profile_api->profileFindOne(json_encode($profile_filter));
+    }
+    catch (ApiException $e) {
+      \Drupal::logger('DBCDK Community Service')->error($e);
```
kasperg: I tend to retrieve the logger factory from the container and inject the logger instance into the class to reduce coupling.

What do you think about this?

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
```diff
@@ -0,0 +1,177 @@
...
+  public static function create(ContainerInterface $container) {
+    try {
+      $profile_filter = [
+        'where' => [
+          'username' => $container->get('request_stack')->getCurrentRequest()->get('username'),
+        ],
+      ];
+      /* @var \DBCDK\CommunityServices\Api\ProfileApi */
+      $profile_api = $container->get('dbcdk_community.api.profile');
+      $profile = $profile_api->profileFindOne(json_encode($profile_filter));
```
kasperg: This is the create form. Why do would we try to retrieve an existing user?

web/modules/custom/dbcdk_community/src/Form/QuarantineEditForm.php
```diff
@@ -0,0 +1,203 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\dbcdk_community\Form\QuarantineEditForm.
```
kasperg: There seems to be a lot of duplication here between the create and edit form. How about reducing this to a single form based on the following principles:
1. `create` fetches a `Quarantine` from the API if a request argument is present or creates an empty one if not
2. The empty or populated `Quarantine` is passed to the form constructor
3. `buildForm` creates the form with default values from the `Quarantine` object and [stores the `Quarantine` in the form state](https://github.com/DBCDK/biblo-admin/pull/39/files#diff-43b66a6be6f52cc9f63ae814391bcc2eR82).
4. [`submitForm` retrieves the `Quarantine` from the form state](https://github.com/DBCDK/biblo-admin/pull/39/files#diff-43b66a6be6f52cc9f63ae814391bcc2eR98), sets the values and calls create or upsert on the API based on whether the `Quarantine` has an id or not.

web/modules/custom/dbcdk_community/src/Form/QuarantineEditForm.php
```diff
@@ -0,0 +1,203 @@
...
+      $quarantine = $quarantine_api->quarantineFindOne(json_encode($quarantine_filter));
+
+      $profile_filter = [
+        'where' => [
+          'id' => $quarantine->getQuarantinedId(),
+        ],
+      ];
+      /* @var \DBCDK\CommunityServices\Api\ProfileApi */
+      $profile_api = $container->get('dbcdk_community.api.profile');
+      $profile = $profile_api->profileFindOne(json_encode($profile_filter));
```
kasperg: How about using [`profileFindById`](https://github.com/DBCDK/biblo-admin/blob/f1abf0227666af4a883b46cace763fb6ca7ad407/lib/client/src/Api/ProfileApi.php#L1184-L1184) here?

web/modules/custom/dbcdk_community/src/Form/QuarantineEditForm.php
```diff
@@ -0,0 +1,203 @@
...
+  public static function create(ContainerInterface $container) {
+    try {
+      /* @var \DBCDK\CommunityServices\Api\QuarantineApi */
+      $quarantine_api = $container->get('dbcdk_community.api.quarantine');
+      $quarantine_filter = [
+        'where' => [
+          'id' => $container->get('request_stack')->getCurrentRequest()->get('quarantine_id'),
+        ],
+      ];
+      $quarantine = $quarantine_api->quarantineFindOne(json_encode($quarantine_filter));
```
kasperg: There is also [`quarantineFindById`](https://github.com/DBCDK/biblo-admin/blob/f1abf0227666af4a883b46cace763fb6ca7ad407/lib/client/src/Api/QuarantineApi.php#L800-L800).

web/config/sync/core.date_format.dbcdk_community_service_date.yml
```diff
@@ -0,0 +1,8 @@
+uuid: 85e1d2d9-09e5-41de-8900-a8a9c4486a8b
+langcode: en
+status: true
+dependencies: {  }
+id: dbcdk_community_service_date
+label: 'DBCDK Community Service date'
+locked: false
+pattern: 'j. F Y'
```
kasperg, line 8: :+1:

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * @Block(
+ *   id = "dbcdk_community_profile_quarantines_block",
+ *   admin_label = @Translation("DBCDK Community Profile Quarantines"),
+ *   context = {
+ *     "username" = @ContextDefinition("string")
+ *   }
+ * )
+ */
+class ProfileQuarantinesBlock extends BlockBase implements ContainerFactoryPluginInterface {
```
kasperg, line 35: IMO the `Block` part is duplicated information (it is already indicated in the namespace) but that may be a matter of preference.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
```diff
@@ -0,0 +1,213 @@
...
+      $container->get('dbcdk_community.api.quarantine'),
+      $container->get('date.formatter')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    $context_username = $this->getContext('username')->getContextData()->getValue();
```
kasperg: How about using `getContextValue()` instead of the chained calls?

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
```diff
@@ -0,0 +1,213 @@
...
+    $quarantines = [];
+    try {
+      // Fetch a Community Profile based on the username so we can the
+      // corresponding ID to fetch quarantines.
+      $profile_filter = [
+        'where' => [
+          'username' => $context_username,
+        ],
+      ];
+      /* @var \DBCDK\CommunityServices\Model\Profile */
```
kasperg: Is this type hint necessary? [It is already part of the DocBlock for `profileFindOne`](https://github.com/DBCDK/biblo-admin/blob/f1abf0227666af4a883b46cace763fb6ca7ad407/lib/client/src/Api/ProfileApi.php#L802-L802).

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
```diff
@@ -0,0 +1,213 @@
...
+
+      // Fetch quarantines.
+      $quarantine_filter = [
+        'order' => 'end DESC',
+        'where' => [
+          'quarantinedId' => $profile->getId(),
+        ],
+      ];
+      /* @var \DBCDK\CommunityServices\Model\Quarantine[] */
+      $quarantines = $this->quarantineApi->quarantineFind(json_encode($quarantine_filter));
```
kasperg: Again: Is the typehint necessary? Also we need to cast to array to ensure handling of empty responses which are returned as `NULL`.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
```diff
@@ -0,0 +1,213 @@
...
+
+    // Build table of quarantines.
+    $build = [];
+    $build['table'] = [
+      '#theme' => 'table',
+      '#header' => $fields,
+      '#rows' => $rows,
+      '#empty' => $this->t('There are no quarantines attached to this profile.', [
+        '%username' => $context_username,
+      ]),
```
kasperg: The `%username` placeholder is not used in the string?

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
kallehauge: Yes - My understanding is that the reason is without markup and "textarea" is equivalent to plain_text :)

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
kallehauge: Yes, if you look at the model from the explorer:

```
Quarantine {
  reason (string, optional),
  end (string),
  start (string),
  id (number, optional),
  quarantinedId (number, optional),
  creatorId (number, optional)
}
```

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
kallehauge: Great idea! :+1:

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * Provides a 'ProfileQuarantinesBlock' block.
+ *
+ * This block provides an overview of a Community Profile's Quarantines.
+ *
+ * @Block(
+ *   id = "dbcdk_community_profile_quarantines_block",
+ *   admin_label = @Translation("DBCDK Community Profile Quarantines"),
+ *   context = {
+ *     "username" = @ContextDefinition("string")
```
kasperg, line 31: Getting Profiles by username is a reoccuring thing but it seems to make some things a bit more complex e.g.
- Gettings profiles where we use `profileFindOne` with filter instead of `profileFindById`
- Generating links to quarantines where we have to pass the username along instead of using the `quarantinedId` already available on the `Quarantine` object
- Are we certain that the characters in usernames are enforced to be suitable url arguments?

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
kasperg: I was unclear. I was thinking that it should be WYSIWYG.

As I understood it the reason field should be WYSIWYG as that allowed moderators to communicate with eathother about the background for the ban, include links to offending content etc.

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
kasperg: :+1:

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
kallehauge: Good catch - it was a leftover from an initial message that was changed during development :)

kasperg:
Reason complete.

A lot of bits and pieces here. I think the primary issues is about form handling/duplication and whether profile urls shoud be based on username or profile id.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
kallehauge, line 35: If you look in Drupal core, then the naming convention seems to be "InsertNameBlock". The same with Forms (InsertNameForm) etc. - but I have not found this in writing anywhere

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
kallehauge: :+1:

kallehauge:
@kasperg the small bits and pieces and duplication should be dealt with now.

In regards to the profile URLs, then I agree with you. It makes sense to switch to IDs instead - but it's a bit out of scope from this PR

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
kallehauge: the DATE_CONSTANT does no longer exist. The forms have been merged to one, but you can find the changes [here](https://github.com/DBCDK/biblo-admin/pull/38/files#diff-f2a84a73f7355c1bbbdd2530354f88e1R122) and in the ::submitForm() of the class.

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
kallehauge: Ah yes! I've [changed the field](https://github.com/DBCDK/biblo-admin/pull/38/files#diff-f2a84a73f7355c1bbbdd2530354f88e1R115) to use the "Comunity basic markup" text format that supports: strong, emphasized and links.

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
kallehauge: https://github.com/DBCDK/biblo-admin/pull/38/files#diff-f2a84a73f7355c1bbbdd2530354f88e1R188

web/modules/custom/dbcdk_community/src/Form/QuarantineCreateForm.php
kallehauge: https://github.com/DBCDK/biblo-admin/pull/38/files#diff-f2a84a73f7355c1bbbdd2530354f88e1R166

web/modules/custom/dbcdk_community/src/Form/QuarantineEditForm.php
kallehauge: The two forms have been merged to one but I've not instantiated an empty `Quarantine` in the constructor since it seemed more clean to check the ::buildForm if "Quarantine" was empty instead of checking for empty values in the empty `Quarantine` - I do it in [::submitForm() instead](https://github.com/DBCDK/biblo-admin/pull/38/files#diff-f2a84a73f7355c1bbbdd2530354f88e1R165).

web/modules/custom/dbcdk_community/src/Form/QuarantineEditForm.php
kallehauge: This has been implemented here: https://github.com/DBCDK/biblo-admin/pull/38/files#diff-f2a84a73f7355c1bbbdd2530354f88e1R72

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
kallehauge: Removed and casted

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
kasperg, line 35: :+1:

kasperg:
@kallehauge: Looks great. You can merge this when you want. Consider rebasing first?

PR #42, opened 2016-04-04
Add the option to filter profiles by quarantines

kallehauge:
This PR contains:
- A new field in the Profiles Filter form.
- A check in Profiles Block that determines how we should fetch the list of profiles.

### DIFF
--- a/web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
+++ b/web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
@@ -8,8 +8,10 @@
 namespace Drupal\dbcdk_community\Plugin\Block;
 
 use DBCDK\CommunityServices\Api\ProfileApi;
+use DBCDK\CommunityServices\Api\QuarantineApi;
 use DBCDK\CommunityServices\ApiException;
 use DBCDK\CommunityServices\Model\Profile;
+use DBCDK\CommunityServices\Model\Quarantine;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\Core\Block\BlockBase;
@@ -43,9 +45,16 @@ class ProfilesBlock extends BlockBase implements ContainerFactoryPluginInterface
   /**
    * The DBCDK Community Service Profile API.
    *
-   * @var ProfileApi $profileApi
+   * @var \DBCDK\CommunityServices\Api\ProfileApi $profileApi
    */
   protected $profileApi;
+
+  /**
+   * The DBCDK Community Service Quarantine API.
+   *
+   * @var \DBCDK\CommunityServices\Api\QuarantineApi $quarantineApi
+   */
+  protected $quarantineApi;
 
   /**
    * Generator to use when creating urls for community content on frontend site.
@@ -76,6 +85,16 @@ class ProfilesBlock extends BlockBase implements ContainerFactoryPluginInterface
   protected $pageNumber;
 
   /**
+   * Quarantined filter state.
+   *
+   * Positive if the list of profiles should be filtered by profiles with an
+   * active quarantine.
+   *
+   * @var int $quarantinedFilter
+   */
+  protected $quarantinedFilter;
+
+  /**
    * Creates a Profiles Block instance.
    *
    * @param array $configuration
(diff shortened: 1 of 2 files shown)


web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
```diff
@@ -76,6 +84,16 @@ class ProfilesBlock extends BlockBase implements ContainerFactoryPluginInterface
   protected $pageNumber;
 
   /**
+   * Quarantined filter state.
+   *
+   * Positive if the list of profiles should be filtered by profiles with an
+   * active quarantine.
+   *
+   * @var int $quarantinedFilter
```
kasperg, line 93: How about making this a boolean instead of an integer?

That seems to be more appropriate for what the filter actually does.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
```diff
@@ -167,21 +193,37 @@ public function build() {
...
-      $profiles = (array) $this->profileApi->profileFind(json_encode($filter));
+      // Fetch a list or profiles with an active quarantine.
+      if ($this->quarantinedFilter) {
+        // The "count" request doesn't work, so we remove the "limit" and
+        // "offset" filter arguments to we can count all the results for the
+        // pager and slice the results ourselves.
+        // @TODO: Remove the "unsets" when we can use the "profileCount()"
+        // method from the swagger generated community service client.
+        unset($filter['limit']);
+        unset($filter['offset']);
```
kasperg, line 212: I need to have this explained in person to get why this is necessary.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
```diff
@@ -208,6 +250,48 @@ public function build() {
...
+   *
+   * @return \DBCDK\CommunityServices\Model\Profile[] $profiles
+   *   An array of Profile objects with an active quarantine.
+   */
+  protected function getQuarantinedProfiles(array $profiles_filter) {
+    // Fetch all quarantines that are active from the moment of the request.
+    $quarantined_filter = [
+      'where' => [
+        'end' => [
+          'gt' => date('Y-m-d\TH:i:s.000\Z'),
```
kasperg: Could we use one of [the predefined formats](http://php.net/manual/en/class.datetime.php#datetime.constants.types) here?

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
```diff
@@ -208,6 +250,48 @@ public function build() {
...
+      ],
+    ];
+    $quarantines = (array) $this->quarantineApi->quarantineFind(json_encode($quarantined_filter));
+
+    // Reduce all the quarantines to an array of quarantined profile ids.
+    // We do this to make sure a profile id only appears once since we use these
+    // values as "Where" arguments for the next request to the service, and a
+    // profile can have multiple quarantines at the same time.
+    $quarantine_ids = array_reduce($quarantines, function($result, $quarantine) {
+      $result[$quarantine->getQuarantinedId()] = ['id' => $quarantine->getQuarantinedId()];
```
kasperg: We need to update this according to the latest changes in #48.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
```diff
@@ -208,6 +250,48 @@ public function build() {
...
+      return $result;
+    });
+
+    // Reset array keys so json_encode() will handle the ids as an indexed array
+    // and not an associative array.
+    $quarantine_ids = array_values($quarantine_ids);
```
kasperg, line 290: Why do we set array keys in L280 only to strip them 6 lines below?

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
```diff
@@ -208,6 +250,48 @@ public function build() {
...
+   * @param array $profiles_filter
+   *   An array containing filter arguments.
+   *
+   * @throws \DBCDK\CommunityServices\ApiException
+   *   Throws API Exception if any of the calls to the services fails.
+   *
+   * @return \DBCDK\CommunityServices\Model\Profile[] $profiles
+   *   An array of Profile objects with an active quarantine.
+   */
+  protected function getQuarantinedProfiles(array $profiles_filter) {
```
kasperg, line 265: Methods like this point in the direction of a separate ProfileRepository class which can encapsulate such logic. Perhaps later on though.

kasperg:
Review complete. Lets go this through together.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
kallehauge, line 290: We set the array keys to remove any duplications of user id's compared to just pushing to `$result[] =` - it could be discussed if this is necessary and if a user would have two active quarantines at the same time, but it's defensive coding to counter that scenario (I could be convinced to NOT do this if we can agree that multiple "where id = ..." with the same user isn't a real issue here

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
kallehauge, line 265: You're 100% correct :+1:

PR #47, opened 2016-04-04
Refactor "Edit Profile" form

kallehauge:
- Handle dates by using \DateTime().
- Allow for NULL values for $profile in the constructor.
- Remove the complexity of checking if there have been made changes to any values.
- Pass the Profile model instead of an array of data when updating a profile.

### DIFF
--- a/web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
+++ b/web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
@@ -10,21 +10,17 @@ namespace Drupal\dbcdk_community\Form;
 use DBCDK\CommunityServices\ApiException;
 use DBCDK\CommunityServices\Api\ProfileApi;
 use DBCDK\CommunityServices\Model\Profile;
-use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\Core\Datetime\DrupalDateTime;
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
 use Drupal\Core\Form\FormBase;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Url;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Edit a Community Service Profile.
  */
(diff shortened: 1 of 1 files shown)


web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -153,10 +149,10 @@ public function buildForm(array $form, FormStateInterface $form_state, $username
     ];
 
     $form['birthday'] = [
-      '#type' => 'date',
+      '#type' => 'datetime',
```
kasperg, line 153: Why is using a `datetime` element better than using a `date` for representing birthdays in a form?

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -225,105 +221,69 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
    * @throws \DBCDK\CommunityServices\ApiException
    *   Throws API Exception if the roles could not be added/removed.
-   *
-   * @return bool
-   *   Returns TRUE if there were any changes to update and vise versa.
    */
   protected function updateProfileRoles(array $new_role_ids) {
     // Format an array of the profile's role ids before any update has happened.
     $old_role_ids = array_map(function($role) {
+      /* @var \DBCDK\CommunityServices\Model\CommunityRole $role */
```
kasperg: How about typing hinting `$role` to `CommunityRole` instead in the closure?

kasperg:
@kallehauge: Review complete. Looks good. I have a couple of questions and I think the Scrutinizer errors can be ignored.

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
kallehauge, line 153: It's not necessarily better, it's just cleaner since we don't have to use a [class constant](https://github.com/DBCDK/biblo-admin/pull/47/files#diff-3fb04a6aef57ae4c52468bf184cacdebL27)

PR #48, opened 2016-04-04
Community client update

kasperg:
This updates the client library with the latest changes to the community service.

The surrounding code is updated accordingly.

### DIFF
--- a/web/modules/custom/dbcdk_community/src/Content/FlaggableContentRepository.php
+++ b/web/modules/custom/dbcdk_community/src/Content/FlaggableContentRepository.php
@@ -90,6 +90,7 @@ class FlaggableContentRepository {
     // The API cannot handle an empty array being passed so we convert it to
     // an empty string.
     $filter_string = (!empty($filter)) ? json_encode($filter) : NULL;
+    /* @var \DBCDK\CommunityServices\Model\Flag[] $flags */
     $flags = (array) $this->flagApi->flagFind($filter_string);
 
     // Loop over *all* flags to find which content it is attached to.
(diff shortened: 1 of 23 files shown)


kasperg:
I will merge this to people can actually test.

Please review this afterwards @kallehauge.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
```diff
@@ -108,14 +108,7 @@ public function build() {
...
 
-      // Fetch quarantines.
-      $quarantine_filter = [
-        'order' => 'end DESC',
-        'where' => [
-          'quarantinedId' => $profile->getId(),
-        ],
-      ];
-      $quarantines = (array) $this->quarantineApi->quarantineFind(json_encode($quarantine_filter));
+      $quarantines = (array) $this->profileApi->profilePrototypeGetQuarantines($profile->getId(), json_encode(['order' => 'end DESC']));
```
kallehauge, line 111: I might not completely agree with using Prototype methods where we don't necessarily have to since they are declared as a "state of not done" (or that is my understanding of it) - but since we're manually updating the Community Service Client, then this will not cause any sudden problems so I don't mind it

web/modules/custom/dbcdk_community/src/Content/FlaggableContentRepository.php
```diff
@@ -100,20 +101,13 @@ protected function getContent(array $filter = []) {
       // We do not know which type of content our flag is attached to and
       // in theory it could be attached to multiple. We get all of them and
       // convert them to FlaggableContent instances.
-      try {
-        $posts = (array) $this->flagApi->flagPrototypeGetPosts($flag->getId());
-        $comments = (array) $this->flagApi->flagPrototypeGetComments($flag->getId());
-      }
-      catch (ApiException $e) {
```
kallehauge, line 107: @kasperg it doesn't look like ApiException is being used anymore. Should we remove it as a "use" and declare the fully qualified namespace for `@throws` in the getContentWithUnreadFlags() DocBlock?

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
kallehauge: Wups, forgot to ping: @kasperg

kallehauge:
I looks good @kasperg. I have a couple of questions and it would have been nice to have the update of the client code in a separate commit - but it's all nitpicking that can be taken care of later on :+1:

PR #55, opened 2016-04-05
Use development settings in Docker environment

kasperg:
We use Docker for development. We could/should automatically configure
the Drupal environment accordingly.

### DIFF
--- a/docker/docker.settings.php
+++ b/docker/docker.settings.php
@@ -18,3 +18,7 @@ $settings['update_free_access'] = FALSE;
 $settings['container_yamls'][] = __DIR__ . '/services.yml';
 
 $settings['install_profile'] = 'minimal';
+
+if (file_exists(__DIR__ . '/settings.local.php')) {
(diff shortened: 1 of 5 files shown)


docker/docker.settings.php
```diff
@@ -18,3 +18,12 @@
 $settings['container_yamls'][] = __DIR__ . '/services.yml';
 
 $settings['install_profile'] = 'minimal';
+
+// Development specific configuration
+// Load Drupal default development configuration
+require_once DRUPAL_ROOT . '/sites/example.settings.local.php';
```
kallehauge: It seems wrong to load the example file? We should keep a settings.local.php file that we mount inside the `/sites/` directory or similar.

kallehauge:
The additions are more than welcome :+1: ! But I have a single question regarding the settings.local.php file we require

kasperg:
@kallehauge: Here is a new attempt.

PR #66, opened 2016-04-11
Fix profile count

kasperg:
The generated client library based on Swagger Codegen 2.1.5 was unable to properly deserialize objects without a defined model class.

[This problem has been fixed subsequently](ttps://github.com/swagger-api/swagger-codegen/pull/2316) and this PR:
- Adds a patched version of Swagger Codegen which fixes that problem
- Updates the generated library based on the update
- Updated the code to use the proper count methods. See 21bcf313bf25aee4f41b6b2e67165106cc8de1a5.


web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
```diff
@@ -12,6 +12,7 @@
 use DBCDK\CommunityServices\ApiException;
 use DBCDK\CommunityServices\Model\Profile;
 use DBCDK\CommunityServices\Model\Quarantine;
+use Psr\Log\NullLogger;
```
kallehauge: I cannot find any usage of this. Is this committed on purpose @kasperg ?

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
```diff
@@ -301,6 +283,52 @@ protected function getQuarantinedProfiles(array $profiles_filter) {
...
+
+    // Reset array keys so json_encode() will handle the ids as an indexed array
+    // and not an associative array.
+    $quarantine_ids = array_values($quarantine_ids);
+
+    // Use the array of quarantined ids as "where" arguments.
+    $profiles_filter['or'] = $quarantine_ids;
+
+    $result = $this->profileApi->profileCount(json_encode($profiles_filter));
+    return (!empty($result['count'])) ? $result['count'] : NULL;
```
kallehauge: Why do we return NULL instead of 0? It would keep the return type an integer at all times instead of a mixture

kallehauge:
Awesome work getting the count methods to work! :+1: I only have a couple of simple questions.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
kasperg: I did it to be able to separate between there being 0 active quarantines and an error situation with an unexpected result.

The surrounding code does not handle this `NULL` value though so perhaps it is not worth it.

kasperg: Fixed now.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfilesBlock.php
kasperg: Removed. Thanks.

kasperg:
This should be ready for merge now.

PR #67, opened 2016-04-11
Refactor profiles

kasperg:
This change refactors the way the relationship between profiles, roles and quarantines are handled.
- Access to data is refactored into a ProfileRepository class
- Handling of quarantines and community roles is refactored into traits
- Unit tests for repository is added


web/modules/custom/dbcdk_community/src/Profile/HasCommunityRolesTrait.php
```diff
@@ -0,0 +1,42 @@
...
+   *   Related community files.
+   */
+  public function getCommunityRoles() {
+    return $this->communityRoles;
+  }
+
+  /**
+   * Set the community roles related to this object.
+   *
+   * @param |DBCDK\CommunityServices\Model\CommunityRole[] $community_roles
```
kallehauge: Scrutinizer caught this: the namespace should start with a backslash, not a pipe

web/modules/custom/dbcdk_community/src/Profile/HasQuarantinesTrait.php
```diff
@@ -0,0 +1,41 @@
...
+   *   Related quarantines.
+   */
+  public function getQuarantines() {
+    return $this->quarantines;
+  }
+
+  /**
+   * Set quarantines related to this object.
+   *
+   * @param |DBCDK\CommunityServices\Model\Quarantine[] $quarantines
```
kallehauge: Scrutinizer caught this: the namespace should start with a backslash, not a pipe

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -24,79 +24,50 @@
 class ProfileEditForm extends FormBase implements ContainerInjectionInterface {
 
   /**
-   * The DBCDK Community Service Profile API.
+   * The profile repository to use.
    *
-   * @var \DBCDK\CommunityServices\Api\ProfileApi $profileApi
+   * @var \Drupal\dbcdk_community\Profile\ProfileRepository $profileRepository
```
kallehauge: The ProfileRepository is declared at line 18, but you use the full name here - at line 36 it looks like you removed the namespace of the object `@var Profile $profile` - It is without any functional effect but my OCD thinks it would be nice if it was consistent :smile:

web/modules/custom/dbcdk_community/src/Form/ProfileEditForm.php
```diff
@@ -24,79 +24,50 @@
...
-
-  /**
    * Creates a Profile Edit Form instance.
    *
-   * @param \DBCDK\CommunityServices\Api\ProfileApi $profile_api
-   *   The DBCDK Community Service Profile API.
-   * @param \DBCDK\CommunityServices\Model\Profile $profile
+   * @param \Drupal\dbcdk_community\Profile\ProfileRepository $profile_repository
+   *   The profile repository to use.
+   * @param \DBCDK\CommunityServices\Model\Profile|\Drupal\dbcdk_community\Profile\Profile $profile
```
kallehauge: When I glans through the code in this class, it assumes that it's a `\Drupal\dbcdk_community\Profile\Profile` that has been passed (ex: it assumes that it can fetch roles with the [`Profile::getCommunityRoles`](https://github.com/DBCDK/biblo-admin/pull/67/files#diff-3fb04a6aef57ae4c52468bf184cacdebR135) method. And from what I can see, then the Repository will always [return an "enriched" profile or an empty result](https://github.com/DBCDK/biblo-admin/blob/feature/refactor-profiles/web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php#L136).

If this is the case and the `\DBCDK\CommunityServices\Model\Profile` is removed from the documentation, then I feel like the `protected $profile` property should have a fully declared namespace documented as well (`\Drupal\dbcdk_community\Profile\Profile`).

kasperg: You are totally right. It should always be a subclassed/enriched Profile passed as an argument.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
```diff
@@ -75,14 +76,10 @@ public function build() {
...
     try {
-      $filter = [
-        'where' => [
-          'username' => $username,
-        ],
-      ];
-      $profile = $this->profileApi->profileFindOne(json_encode($filter));
+      $profile = $this->profileRepository->getProfileByUsername(
+        $this->getContextValue('username')
+      );
```
kallehauge, line 89: @kasperg did you split this into 3 lines because Scrutinizer complained about the length of the call?

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileQuarantinesBlock.php
```diff
@@ -96,23 +85,14 @@ public static function create(ContainerInterface $container, array $configuratio
...
-
-      $quarantines = (array) $this->profileApi->profilePrototypeGetQuarantines($profile->getId(), json_encode(['order' => 'end DESC']));
+      $profile = $this->profileRepository->getProfileByUsername($this->getContextValue('username'));
     }
     catch (ApiException $e) {
       \Drupal::logger('DBCDK Community Service')->error($e);
-      $profile = NULL;
+      $profile = new Profile();
+      // The generated Profile model does not have a public quarantines
+      // attribute but the repository will add it when retrieving the object.
```
kallehauge: Is this a misplaced comment or leftover from earlier code that did not make the cut at the end?

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
kasperg, line 89: To be honest I think PhpStorm did it for me automatically when editing the line.

I am not sure PHPCS would actually complain about it.

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
```diff
@@ -0,0 +1,357 @@
+<?php
+/**
+ * @file
+ * Class definition for ProfileRepository.
+ */
+
+namespace Drupal\dbcdk_community\Profile;
+
+
```
kallehauge: 2x new line (OCD correction)

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
```diff
@@ -0,0 +1,357 @@
...
+   * @param \DBCDK\CommunityServices\Api\QuarantineApi $quarantine_api
+   *   The quarantine api to use.
+   * @param \DBCDK\CommunityServices\Api\CommunityRoleApi $community_role_api
+   *   The community role api to use.
+   */
+  public function __construct(ProfileApi $profile_api, QuarantineApi $quarantine_api, CommunityRoleApi $community_role_api) {
+    $this->profileApi = $profile_api;
+    $this->quarantineApi = $quarantine_api;
+    $this->communityRoleApi = $community_role_api;
+    $this->logger = new NullLogger();
```
kallehauge, line 62: Should there be a comment here explaining that we define the NullLogger inside LoggerAwareTrait?

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
```diff
@@ -0,0 +1,357 @@
...
+   * @param array $filter
+   *   The filter. This can contain entries such as where, limit, offset.
+   *
+   * @return string
+   *   A JSON encoded version of the filter.
+   *
+   * @see https://docs.strongloop.com/display/public/LB/Querying+data
+   */
+  protected function processFilter(array $filter) {
+    return (!empty($filter)) ? json_encode($filter) : NULL;
```
kallehauge: According to the DocBlock, this should return a string. Either the return should be changed or the DocBlock needs an update :relaxed:

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
```diff
@@ -0,0 +1,357 @@
...
+   */
+  public function getProfileByUsername($username) {
+    $filter = [
+      'where' => ['username' => $username],
+    ];
+    return $this->findProfile($filter);
+  }
+
+  /**
+   * Retrieve a profile matching a filter with community roles and quarantines.
```
kallehauge: The basic functionality of this method is to: Find a profile.

That it contains CommunityRoles and Quarantines, shouldn't that be part of an expanded description? Example:

```
/**
 * Retrieve a profile matching a filter.
 *
 * The retrieved profile will be enriched with community roles and quarantines. We do this because...
...
```

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
```diff
@@ -0,0 +1,357 @@
...
+  /**
+   * Enrich a "flat" profile object.
+   *
+   * This adds details about community roles and quarantines.
+   *
+   * @param ModelProfile $profile
+   *   The original profile object.
+   *
+   * @return Profile
+   *   The profile object enriched with details ab
```
kallehauge: This comment seems a bit "unfinished" :smile:

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
```diff
@@ -0,0 +1,357 @@
...
+
+  /**
+   * Get profiles with an active quarantine at a specific time.
+   *
+   * @param DateTime $time
+   *   The time to check quarantine date ranges against.
+   * @param array $profiles_filter
+   *   An array containing filter arguments.
+   *
+   * @throws \DBCDK\CommunityServices\ApiException
```
kallehauge: In most (if not all) of our DocBlocks we document `@throws` below `@return`. An example would be the method below this one (countQuarantinedProfiles). CodeSniffer and Scrutinizer will not complain about this - but my OCD kicks in a bit :smiley:

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
```diff
@@ -0,0 +1,357 @@
...
+
+  /**
+   * Count the number of quarantined profiles matching a filter.
+   *
+   * @param DateTime $time
+   *   The time to check quarantine date ranges against.
+   * @param array $profiles_filter
+   *   The filter to match profiles against.
+   *
+   * @return int
```
kallehauge: Shouldn't this be `int|NULL`?

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
```diff
@@ -0,0 +1,357 @@
...
+
+    // Use the array of quarantined ids as "where" arguments.
+    $profiles_filter['where']['or'] = $this->getQuarantinedProfilesFilter($quarantines);
+
+    return (array) $this->profileApi->profileFind($this->processFilter($profiles_filter));
+  }
+
+  /**
+   * Count the number of quarantined profiles matching a filter.
+   *
```
kallehauge, line 236: Should we include a description saying that it's not a simple "count" request but that it actually fetches all profiles and then counts them?

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
```diff
@@ -0,0 +1,357 @@
...
+
+    $result = $this->profileApi->profileCount($this->processFilter($profiles_filter));
+    return (isset($result['count'])) ? $result['count'] : NULL;
+  }
+
+  /**
+   * Get all active quarantines.
+   *
+   * An active quarantine is a quarantine where the end date has not
+   * yet passed.
```
kallehauge, line 263: :+1:

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
```diff
@@ -0,0 +1,357 @@
...
+  }
+
+  /**
+   * Build an array of quarantined profile ids suitable for API filtering.
+   *
+   * @param Quarantine[] $quarantines
+   *   The quarantines to get profiles from.
+   *
+   * @return array
+   *   An filter array.
```
kallehauge: Typo:

> A filter array.

web/modules/custom/dbcdk_community/tests/src/Unit/Profile/ProfileRepositoryTest.php
```diff
@@ -0,0 +1,288 @@
...
+ */
+class ProfileRepositoryTest extends PHPUnit_Framework_TestCase {
+
+  /**
+   * Test profile retrieval by id.
+   */
+  public function testProfileById() {
+    $profile = (new ModelProfile())->setId(1);
+    $profile_api = $this->getMock('\DBCDK\CommunityServices\Api\ProfileApi');
+    $profile_api->method('profileFindOne')->willReturnMap([
```
kallehauge, line 29: It makes so much sense to use `willReturnMap` here :+1:

web/modules/custom/dbcdk_community/tests/src/Unit/Profile/ProfileRepositoryTest.php
```diff
@@ -0,0 +1,288 @@
...
+      new Profile($profile),
+      $repo->getProfile(1),
+      'Repository does not return expected profile by id'
+    );
+  }
+
+  /**
+   * Test profile retrival with no matching id.
+   */
+  public function testProfileByNonexistingId() {
```
kallehauge: I'm not quite sure what case we're testing here? Is it to make sure that [ProfileRepository::findProfile](https://github.com/DBCDK/biblo-admin/pull/67/files#diff-2b1d161e27289b87681dff9a344e3d49R131) does NOT enrich empty results?

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
kasperg, line 62: I do not think that makes much sense. The fact that it exists is defined when using the trait. I would not think about setting variables of a parent class either.

Would you prefer this alternate version?

``` php
$this->setLogger(new NullLogger());
```

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
kasperg, line 236: I do not understand your comment. It is a simple count of profiles.

web/modules/custom/dbcdk_community/src/Profile/ProfileRepository.php
kallehauge, line 62: Nah, you're right. Keep it as it is :+1:

web/modules/custom/dbcdk_community/tests/src/Unit/Profile/ProfileRepositoryTest.php
kasperg: You are right. We are only testing that PHPUnit mocks return NULL. I removed the test case.

web/modules/custom/dbcdk_community/src/Plugin/Block/ProfileBlock.php
kallehauge, line 89: I believe you're right and it's not anything you should change - I just saw an irregularity from the way you normally write your code (I know you!!) :smile:

kasperg:
I'll close this in favour of #77 which contains both refactoring and tests.

PR #70, opened 2016-04-12
Moderation UI

kallehauge:
This PR tris to make the moderation experience better by:
- Make "Moderation" its own section in the toolbar.
- Hide the "Back to site" button in the toolbar since we don't have a "site" to go back to.

### DIFF
--- a/web/modules/custom/dbcdk_community/dbcdk_community.module
+++ b/web/modules/custom/dbcdk_community/dbcdk_community.module
@@ -30,3 +30,14 @@ function dbcdk_community_theme($existing, $type, $theme, $path) {
     ],
   ];
 }
+
+/**
+ * Implements hook_element_info_alter().
+ *
+ * Add our own library to the toolbar so we can define icons for custom links.
+ */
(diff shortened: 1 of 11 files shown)


web/modules/custom/dbcdk_community/dbcdk_community.module
```diff
@@ -30,3 +30,26 @@ function dbcdk_community_theme($existing, $type, $theme, $path) {
...
+  if (isset($types['toolbar'])) {
+    $types['toolbar']['#attached']['library'][] = 'dbcdk_community/toolbar';
+  }
+}
+
+/**
+ * Implements hook_toolbar_alter().
+ *
+ * Remove the "home" item since we don't have any front-end to direct the user
+ * back to.
```
kasperg, line 49: :+1:

kasperg:
Looks great to me.

Please also remove the [`access content overview` permission](https://github.com/DBCDK/biblo-admin/blob/develop/web/config/sync/user.role.admin_moderator.yml#L14) from moderators. They do not need it anymore.

kasperg:
:+1:

PR #75, opened 2016-04-18
Fix flagged content list pager

kasperg:
Also add more unit tests to make sure that it does not happen again.

[PHPMock](https://github.com/php-mock/php-mock-phpunit) makes it possible to mock static Drupal calls which are not migrated to services yet.

---

This change is [Reviewable](https://reviewable.io/reviews/dbcdk/biblo-admin/75)


kallehauge:
This is great! I've corrected a few typos in comments but that was it 👍

PR #76, opened 2016-04-19
Feature/profiles list filters

kallehauge:
This PR does three things:
1. Generalise the "open" state of `ProfilesFilterForm`.
2. Add additional optional filter conditions.
3. Make the quarantine filter option work with additional `where` conditions.


kallehauge:
Scrutinizer failed on two "this might be duplicated code" - I've ignored them both so see the build as successful :smile:

kasperg:
@kallehauge Looks good.

I have introduced 2e1d602 to reduce the filtering complexity (and fix the duplicated code).

PR #83, opened 2016-05-02
Separate community module

kallehauge:
This PR is build on PR #82.

This pull-request separates the community module into 3 modules:
1. Base module (the "old" module named dbcdk_community)
2. Content module
3. Moderation module

The content module is currently "empty" since we have only worked with base setup and moderation so far.


kasperg:
Great!

PR #84, opened 2016-05-02
#543: Review moderation

kasperg:
This PR adds support for moderating reviews which have been flagged by users.
- The community service client is updated to add the `Review` model and API
- Existing flaggable content class is refactored to use subclasses of generated models with interfaces and traits. This allows for variation in individual models instead.
- Added support for showing flagged reviews in list


kasperg:
@kallehauge: This is ready for review.

Scrutinizer complains about code duplication. It is true that there is some duplication but I do not see it as a problem nor do I see a good way to change it.

web/modules/custom/dbcdk_community_moderation/src/Content/FlaggableTrait.php
```diff
@@ -0,0 +1,105 @@
+<?php
+/**
+ * @file
+ * FlaggableTrait definition.
+ */
+
+namespace Drupal\dbcdk_community_moderation\Content;
+
+
+use DBCDK\CommunityServices\Model\Flag;
```
kallehauge, line 9: Double new line above `use`

kallehauge:
It looks good! :+1: I found a double new-line but that was it.
I have ignored the issue in Scrutinizer so the last failed build can be overlooked. You can merge at will @kasperg :relaxed:

PR #86, opened 2016-05-03
Add Article content type

kallehauge:
This PR is basically a huge stack of configuration-files doing the following:
- Update `composer.json` to include _Paragraphs_ and _Video_ modules.
- Enable the following modules:
  - Paragraphs (and entity_reference_revisions).
  - Field UI.
  - Taxonomy.
  - Image.
  - Video.
- Create a new content type called _Article_ with fields:
  - Title.
  - Article type (taxonomy ref).
  - Author (textfield).
  - Content (paragraphs ref).
- Create a taxonomy called _Article types_.
- Add Paragraphs types:
  - Text: contains a textarea field.
  - Image: contains an image field.
  - Embedded Video: contains a video field for embed url.

PR #87, opened 2016-05-03
Editor role

kallehauge:
This PR is based upon PR #86.

Add an Editor role with basic rights to create/edit/delete Articles.

### DIFF
--- a/web/config/sync/user.role.admin_editor.yml
+++ b/web/config/sync/user.role.admin_editor.yml
@@ -1,0 +1,21 @@
+uuid: 7e4afd72-7492-499a-a81e-303c00f4ace7
+langcode: en
+status: true
+dependencies: {  }
+id: admin_editor
+label: Editor
+weight: 4
+is_admin: null
+permissions:
+  - 'access content overview'
+  - 'delete all revisions'
+  - 'revert all revisions'
+  - 'view all revisions'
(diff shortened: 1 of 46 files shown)


web/config/sync/user.role.authenticated.yml
```diff
@@ -12,3 +12,4 @@ permissions:
   0: 'access content'
   2: 'view own unpublished content'
   4: 'access toolbar'
+  5: 'view the administration theme'
```
kasperg, line 15: If we add this permission to authenticated can we then remove it from other roles as well?

web/config/sync/user.role.admin_editor.yml
```diff
@@ -0,0 +1,21 @@
...
+  - 'revert all revisions'
+  - 'view all revisions'
+  - 'create article content'
+  - 'delete any article content'
+  - 'delete own article content'
+  - 'delete article revisions'
+  - 'edit any article content'
+  - 'edit own article content'
+  - 'revert article revisions'
+  - 'view article revisions'
```
kasperg, line 21: We have to run these permissions by @AndersFriisBrodsgaard to make sure he agrees - especially the part about own or any.

kasperg:
We will have to show this to him during demo. Reviewed and approved.

PR #90, opened 2016-05-08
Article URLs

kallehauge:
This PR makes it possible to organise articles in a "tree like" URL structure and adds the possibility to customise the slug.

### DIFF
--- a/web/config/sync/core.extension.yml
+++ b/web/config/sync/core.extension.yml
@@ -22,15 +22,18 @@ module:
   page_cache: 0
   page_manager: 0
   paragraphs: 0
+  path: 0
   rest: 0
   serialization: 0
   shortcut: 0
   system: 0
   taxonomy: 0
   text: 0
+  token: 0
   toolbar: 0
   user: 0
   video_embed_field: 0
+  pathauto: 1
   minimal: 1000
 theme:
   stable: 0
--- a/web/modules/custom/dbcdk_community_content/dbcdk_community_content.module
+++ b/web/modules/custom/dbcdk_community_content/dbcdk_community_content.module
@@ -1,0 +1,69 @@
+<?php
+/**
+ * @file
+ * Module file for dbcdk_community_content.
+ */
+
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\node\Entity\Node;
+
+/**
+ * Implements hook_node_presave().
+ *
+ * Duplicate the title of a node to the slug field if the slug field is empty.
+ * We do this because article urls depend on having a slug field value for the
+ * article Pathauto pattern.
+ */
+function dbcdk_community_content_node_presave(Node $node) {
+  if (empty($node->get('field_slug')->getString())) {
+    /* @var \Drupal\pathauto\AliasCleaner $alias_cleaner */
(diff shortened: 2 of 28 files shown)


web/config/sync/core.extension.yml
```diff
@@ -22,15 +22,18 @@ module:
...
   serialization: 0
   shortcut: 0
   system: 0
   taxonomy: 0
   text: 0
+  token: 0
   toolbar: 0
   user: 0
   video_embed_field: 0
+  pathauto: 1
```
kasperg, line 36: Why does this module have the value `1` while the others have `0`?

web/modules/custom/dbcdk_community_content/dbcdk_community_content.module
```diff
@@ -0,0 +1,65 @@
...
+    $node->set('field_slug', $alias_cleaner->cleanString($node->getTitle()));
+  }
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ *
+ * Make the "Slug" and "Sub Path" fields part of an advanced details group.
+ * We do this to give a better editorial experience by placing all additional
+ * options for the content together.
```
kasperg, line 31: :+1:

web/modules/custom/dbcdk_community_content/dbcdk_community_content.tokens.inc
```diff
@@ -0,0 +1,51 @@
...
+ * referenced entities "sub values".
+ */
+function dbcdk_community_content_tokens($type, array $tokens, array $data, array $options, BubbleableMetadata $bubbleable_metadata) {
+  $replacements = array();
+  if ($type === 'entity' & !empty($data['entity'])) {
+    /** @var \Drupal\Core\Entity\EntityInterface $entity */
+    $entity = $data['entity'];
+    foreach ($tokens as $name => $original) {
+      switch ($name) {
+        case 'sub-path':
```
kasperg: How about replacing `foreach/switch/case` with an `if ($tokens['sub-path'])`?

That seems a bit simpler than looping over all tokens.

web/modules/custom/dbcdk_community_content/dbcdk_community_content.module
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Implements hook_node_presave().
+ *
+ * Duplicate the title of a node to the slug field if the slug field is empty.
+ * We do this because article urls depend on having a slug field value for the
+ * article Pathauto pattern.
+ */
+function dbcdk_community_content_node_presave(Node $node) {
+  if (empty($node->get('field_slug')->getString())) {
+    /** @var \Drupal\pathauto\AliasCleaner $alias_cleaner */
```
kasperg: Nitpicking: Double `**` here.

kasperg:
In general I think this looks good. I have a few minor issues here and there.

One thing: We talked about the ability to filter content based on subpath. Will that be added in a later PR? I could not spot an issue for it.

web/config/sync/core.extension.yml
kasperg, line 36: It is the module weight. Never mind :+1:.

kallehauge:
Corrected the issues @kasperg  :relaxed:

Regarding "filter content based on subpath", are you referring to the list of content? If so, then it's a separate ticket that haven't been handlede yet.

PR #91, opened 2016-05-09
Update DBCDK devel dependencies

kallehauge:
Include `field_ui` as a dependency for the DBCDK Community Devel module.
This is required because we've started to work with content types and taxonomies.


kasperg:
Super.

PR #93, opened 2016-05-11
Review list

kasperg:
This PR results in a new page displaying a block showing a list of reviews.
- A new block for displaying an arbitrary from by ID is implemented
- A new form showing a review list with filter is implemented
- A new Page manager page displaying the form in the block is added

We also refactor tests bases:
- The parent test class is also used for the form and thus moved and renamed
- Handling of mock functions is moved to the individual test classes as they are the ones knowing which namespaces the mock functions should exist in

### DIFF
--- a/web/config/sync/page_manager.page_variant.dbcdk_community_reviews.yml
+++ b/web/config/sync/page_manager.page_variant.dbcdk_community_reviews.yml
@@ -1,0 +1,33 @@
+uuid: d7bd6a73-64ef-4f0a-8d4c-7a67c2344733
+langcode: en
+status: true
+dependencies:
+  config:
+    - page_manager.page.dbcdk_community_reviews
+  module:
+    - dbcdk_community
+id: dbcdk_community_reviews
+label: Reviews
+variant: block_display
+variant_settings:
+  blocks:
+    72dda6dc-ddf1-4066-af83-e2a0d9bb5618:
+      id: form_block
+      label: Reviews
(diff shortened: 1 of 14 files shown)


kasperg:
I do not know how to fix the final indentation issue 😭 .

web/config/sync/page_manager.page_variant.dbcdk_community_reviews.yml
```diff
@@ -0,0 +1,33 @@
+uuid: d7bd6a73-64ef-4f0a-8d4c-7a67c2344733
+langcode: en
+status: true
+dependencies:
+  config:
+    - page_manager.page.dbcdk_community_reviews
+  module:
+    - dbcdk_community
+id: dbcdk_community_reviews
```
kallehauge, line 9: +1 for remembering to use the full ID :+1:

web/modules/custom/dbcdk_community_moderation/src/Form/ReviewListForm.php
```diff
@@ -0,0 +1,235 @@
+<?php
+/**
+ * @file
+ * Class definition for review lists.
+ */
+
+namespace Drupal\dbcdk_community_moderation\Form;
+
+
```
kallehauge: Double new line

kallehauge:
I found a double new line and that is it. Great work! :+1:

PR #94, opened 2016-05-17
Disable Drupal's cache for development

kallehauge:
There is pros and cons about this.

Pro: We don't have to cache-rebuild every time we make some changes to ex: anything related to the REST module.

Con: We don't catch if some of our block content is being cached and therefor not updating.


kasperg:
Let's do it.

PR #95, opened 2016-05-17
Expose articles via REST

kallehauge:
This PR contains:
- Add `restui` as part of our devel module
- Update image styles so our exposed images matches biblo.dk's needs.
- Update the REST configuration to only allow [GET] requests.
- Override normalizer services with our own to control the output of the REST requests.
- Core patch to be able to get URL's from file entities.

### DIFF
--- a/web/config/sync/image.style.large.yml
+++ b/web/config/sync/image.style.large.yml
@@ -5,13 +5,13 @@ dependencies: {  }
 _core:
   default_config_hash: J2n0RpFzS0-bgSyxjs6rSdgxB1rb-bTAgqywNx_964M
 name: large
-label: 'Large (480×480)'
+label: 'Large (1024×1024)'
 effects:
   ddd73aa7-4bd6-4c85-b600-bdf2b1628d1d:
     uuid: ddd73aa7-4bd6-4c85-b600-bdf2b1628d1d
     id: image_scale
     weight: 0
     data:
-      width: 480
-      height: 480
+      width: 1024
+      height: 1024
       upscale: false
--- a/web/modules/custom/dbcdk_community_content/src/FieldNormalizer/ImageItemFieldNormalizer.php
+++ b/web/modules/custom/dbcdk_community_content/src/FieldNormalizer/ImageItemFieldNormalizer.php
@@ -1,0 +1,53 @@
+<?php
+
+/**
+ * @file
+ * Contains ImageItemFieldNormalizer.
+ */
+
+namespace Drupal\dbcdk_community_content\FieldNormalizer;
+
+use Drupal\Core\Field\FieldItemBase;
+use Drupal\file\Entity\File;
+use Drupal\image\Entity\ImageStyle;
+
+/**
+ * Normalize the output for an Image field.
+ */
+class ImageItemFieldNormalizer implements FieldNormalizerInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function normalize(FieldItemBase $field) {
+    // Load the file by target_id.
+    /* @var \Drupal\file\Entity\File $image */
+    $image = File::load($field->get('target_id')->getString());
+
+    // Set field values to the output array.
+    $output = [];
+    $output['alt'] = $field->get('alt')->getString();
+    $output['title'] = $field->get('title')->getString();
+    $output['original'] = $image->toUrl()->getUri();
+
+    // Add urls to a processed version of the image.
+    $image_styles = [
+      'thumbnail',
+      'small',
+      'medium',
+      'large',
+    ];
+    foreach ($image_styles as $image_style_name) {
+      /* @var \Drupal\image\Entity\ImageStyle $image_style */
+      $image_style = ImageStyle::load($image_style_name);
+      // ImageStyle::load() will return NULL if the $image_style doesn't exist.
+      if (!empty($image_style)) {
+        // Add the processed image url to the output array.
+        $output[$image_style_name] = $image_style->buildUrl($image->get('uri')->getString());
+      }
+    }
+
+    return $output;
+  }
+
+}
--- a/web/modules/custom/dbcdk_community_content/src/Normalizer/ComplexDataNormalizer.php
+++ b/web/modules/custom/dbcdk_community_content/src/Normalizer/ComplexDataNormalizer.php
@@ -1,0 +1,31 @@
+<?php
+
+/**
+ * @file
+ * Contains ComplexDataNormalizer.
+ */
+
+namespace Drupal\dbcdk_community_content\Normalizer;
+
+use Drupal\serialization\Normalizer\ComplexDataNormalizer as SerializationComplexDataNormalizer;
+
+/**
+ * Normalize Drupal entity object structures.
+ *
+ * This is the default Normalizer for entities. All formats that have Encoders
+ * registered with the Serializer in the DIC will be normalized with this
+ * class unless another Normalizer is registered which supersedes it.
+ *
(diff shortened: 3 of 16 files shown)


web/config/sync/image.style.large.yml
```diff
@@ -5,13 +5,13 @@ dependencies: {  }
 _core:
   default_config_hash: J2n0RpFzS0-bgSyxjs6rSdgxB1rb-bTAgqywNx_964M
 name: large
-label: 'Large (480×480)'
```
kasperg, line 8: These sizes may be good for now but we should talk to the rest of the team whether they fit their use cases.

web/modules/custom/dbcdk_community_devel/dbcdk_community_devel.info.yml
```diff
@@ -9,3 +9,4 @@ dependencies:
   - field_ui
   - page_manager_ui
   - monolog
+  - restui
```
kasperg: Nitpicking: Missing newline at the end of file.

web/modules/custom/dbcdk_community_content/src/Normalizer/EntityReferenceFieldItemNormalizer.php
```diff
@@ -0,0 +1,112 @@
...
+use Drupal\dbcdk_community_content\FieldNormalizer\FieldNormalizer;
+use Drupal\field\Entity\FieldConfig;
+use Drupal\serialization\Normalizer\EntityReferenceFieldItemNormalizer as SerializationEntityReferenceFieldItemNormalizer;
+
+/**
+ * Converts Drupal entity reference item object to an array structure.
+ *
+ * The reason why we override the original EntityReferenceFieldItemNormalizer is
+ * because we wish to expose the value from a referenced entity instead of the
+ * target id.
```
kasperg, line 19: 👍

web/config/sync/image.style.large.yml
kallehauge, line 8: Agreed :+1:

web/modules/custom/dbcdk_community_content/src/FieldNormalizer/ImageItemFieldNormalizer.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+class ImageItemFieldNormalizer implements FieldNormalizerInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function normalize(FieldItemBase $field) {
+    // Load the file by target_id.
+    /* @var \Drupal\file\Entity\File $image */
+    $image = File::load($field->get('target_id')->getString());
```
kasperg, line 25: Could we use a service for this instead?

web/modules/custom/dbcdk_community_content/src/FieldNormalizer/ImageItemFieldNormalizer.php
```diff
@@ -0,0 +1,52 @@
...
+    // Add urls to a processed version of the image.
+    $image_styles = [
+      'thumbnail',
+      'small',
+      'medium',
+      'large',
+    ];
+    foreach ($image_styles as $image_style_name) {
+      /* @var \Drupal\image\Entity\ImageStyle $image_style */
+      $image_style = ImageStyle::load($image_style_name);
```
kasperg, line 42: Could we use a service for this instead?

web/modules/custom/dbcdk_community_content/src/FieldNormalizer/ImageItemFieldNormalizer.php
```diff
@@ -0,0 +1,52 @@
...
+      // ImageStyle::load() will return NULL if the $image_style doesn't exist.
+      if (!empty($image_style)) {
+        // Add the processed image url to the output array.
+        $output[$image_style_name] = $image_style->buildUrl($image->get('uri')->getString());
+      }
```
kasperg, line 47: How about logging a warning instead of failing silently if we try to load a nonexisting imagestyle?

web/modules/custom/dbcdk_community_content/src/Normalizer/ComplexDataNormalizer.php
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Normalize Drupal entity object structures.
+ *
+ * This is the default Normalizer for entities. All formats that have Encoders
+ * registered with the Serializer in the DIC will be normalized with this
+ * class unless another Normalizer is registered which supersedes it.
+ *
+ * We override the original ComplexDataNormalizer to remove the "value" property
+ * name and just return the value.
```
kasperg, line 20: :+1:

web/modules/custom/dbcdk_community_content/src/Normalizer/ContentEntityNormalizer.php
```diff
@@ -0,0 +1,61 @@
...
+        'revision_uid',
+        'revision_log',
+        'revision_translation_affected',
+        'default_langcode',
+        'path',
+        'field_slug',
+        'field_sub_path',
+      ];
+      if (!in_array($name, $blacklist)) {
+        if ($field->access('view', $context['account'])) {
```
kasperg, line 46: :+1: for remembering security here.

web/modules/custom/dbcdk_community_content/src/FieldNormalizer/ImageItemFieldNormalizer.php
kallehauge, line 42: I have not been able to find any. This seems like a similar case to when we have to create links and use `Link::createFromRoute()` (just one example of Link/Url).

web/modules/custom/dbcdk_community_content/src/FieldNormalizer/ImageItemFieldNormalizer.php
kallehauge, line 25: I have not been able to find any. This seems like a similar case to when we have to create links and use Link::createFromRoute() (just one example of Link/Url).

web/modules/custom/dbcdk_community_devel/dbcdk_community_devel.info.yml
kallehauge: Resolved but since it's not a change to line 12, this comment is still showing :)

web/modules/custom/dbcdk_community_content/src/FieldNormalizer/ImageItemFieldNormalizer.php
kasperg, line 25: There difference here is that this will have to access the configuration storage (DB or whatever).

As far as I can tell we have the following options:
1. Change this to [`entity_load('File', $target_id)`](https://api.drupal.org/api/drupal/core%21includes%21entity.inc/function/entity_load/8.1.x). That means we can use php-mock to mock the function call
2. Implement `ContainerFactoryPluginInterface` and inject an [`EntityStorageInterface`](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityStorageInterface.php/interface/EntityStorageInterface/8.1.x) for file entities

web/modules/custom/dbcdk_community_content/src/FieldNormalizer/ImageItemFieldNormalizer.php
kasperg, line 42: As above - but with this I only see that injecting an `EntityStorageInterface` for ImageStyles is an option.

PR #96, opened 2016-05-18
Use Drupal ruleset from Coder module on Scrutinizer

kasperg:
This means we have to configure the analysis as a part of the build -
not as a tool.

See https://scrutinizer-ci.com/docs/tools/php/code-sniffer/


kasperg:
The final Scrutinizer issue can be ignored. This should be good to go.

PR #97, opened 2016-05-23
#541: Review filter

kasperg:
Implement OpenAgency module to interact with service and use service data to present filter options. 

Also use it in other modules instead of library ids.


kallehauge:
Looks awesome! And a great touch that you remembered the generate command :+1:

(PS: Saw your commits to Develop after the merge. :+1:)

PR #98, opened 2016-05-23
#543: New role: Library employee

kasperg:
Library employees are general users who are primarily concerned about
how users related to their library are using the site.

We do not enforce relationships between users and libraries at the 
moment though.

Currently library users can only access the review list so redirect them here 
on login

### DIFF
--- a/web/modules/custom/dbcdk_community_moderation/dbcdk_community_moderation.module
+++ b/web/modules/custom/dbcdk_community_moderation/dbcdk_community_moderation.module
@@ -4,22 +4,6 @@
  * Module file for DBCDK Community Moderation.
  */
 
-use Symfony\Component\HttpFoundation\RedirectResponse;
-use Drupal\Core\Url;
-use Drupal\user\Entity\User;
-
-/**
- * Implements hook_user_login().
- */
-function dbcdk_community_moderation_user_login(User $account) {
-  // Redirect moderators to the flagged content list.
-  if (in_array('admin_moderator', $account->getRoles())) {
-    $url = new Url('page_manager.page_view_dbcdk_community_flagged_content');
-    $redirect = new RedirectResponse($url->toString());
-    $redirect->send();
-  }
-}
-
 /**
  * Implements hook_theme().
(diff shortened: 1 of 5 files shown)


web/modules/custom/dbcdk_community_moderation/dbcdk_community_moderation.module
```diff
@@ -8,9 +8,17 @@
  * Implements hook_user_login().
  */
 function dbcdk_community_moderation_user_login(\Drupal\user\Entity\User $account) {
-  // Redirect moderators to the flagged content list.
-  if (in_array('admin_moderator', $account->getRoles())) {
+  // Redirect users with specific focus to their primary page.
+  // Moderators go to the flagged content page.
+  if ($account->getRoles(TRUE) == ['admin_moderator']) {
```
kallehauge: If I'm correct, then this only assumes that a user has one role. If the user had moderator AND library roles, then they would just hit the user page. Shouldn't we use `in_array`?

kasperg: You are correct. I did this to avoid having to decide which of the redirects is more important.

Do you have a preference?

kallehauge: Hmm. I think it should be moved to the "base" module if the same conditionals should handle possible roles that are both moderation and editor roles (so it's not just the module with the biggest weight that will come first) - Regarding the importances of roles then I believe we should "just" create some by ourselves.
1. Moderation
2. Librarian
3. Editor

I believe this order would cater most to DBC's business case (as many people moderating and reviewing as possible) and then content creation as the last priority since (I assume) this action is the one that will happen the least and will be done by people who understand the system and can easier navigate to their wished location.

kasperg: Fixed and ready for re-review.

kallehauge:
Looks awesome :+1:

PR #99, opened 2016-05-23
#541: More tests for review filters

kasperg:


kasperg:
This is now ready for review.

PR #100, opened 2016-05-25
Add new full HTML text format + editor

kasperg:
Editors want extended capabilities when working with articles. This new
format has more buttons in the editor and in essence allows full HTML.

A few block formats (h1, h5-6) are not allowed as we do not want them to
show up in the editor either.

### DIFF
--- a/web/config/sync/editor.editor.dbcdk_community_full_html.yml
+++ b/web/config/sync/editor.editor.dbcdk_community_full_html.yml
@@ -1,0 +1,50 @@
+uuid: 4491e2ae-de0c-4cc8-b047-ea514595f7cf
+langcode: en
+status: true
+dependencies:
+  config:
+    - filter.format.dbcdk_community_full_html
+  module:
+    - ckeditor
+format: dbcdk_community_full_html
+editor: ckeditor
+settings:
+  toolbar:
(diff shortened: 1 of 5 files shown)


web/config/sync/editor.editor.dbcdk_community_full_html.yml
```diff
@@ -0,0 +1,50 @@
...
+dependencies:
+  config:
+    - filter.format.dbcdk_community_full_html
+  module:
+    - ckeditor
+format: dbcdk_community_full_html
+editor: ckeditor
+settings:
+  toolbar:
+    rows:
```
kallehauge, line 13: These seems fair :+1:

web/config/sync/editor.editor.dbcdk_community_full_html.yml
```diff
@@ -0,0 +1,50 @@
...
+          name: Lists
+          items:
+            - BulletedList
+            - NumberedList
+        -
+          name: Tools
+          items:
+            - Source
+            - RemoveFormat
+            - PasteFromWord
```
kallehauge, line 37: Bah! Copy-paste from Word :smile:

PR #101, opened 2016-05-25
Filter the content list

kallehauge:
Enables the content filter list and alter the default criteria (removed language and added author).

It's not easily possible to add filtration on content since it's paragraphs referenced items. It allows to filter based on target_id etc. but out of the box it's only possible to filter on title.

### DIFF
--- a/web/config/sync/search.page.node_search.yml
+++ b/web/config/sync/search.page.node_search.yml
@@ -1,0 +1,15 @@
+uuid: 82431853-112f-46c2-a5bd-fd4f696ebc23
+langcode: en
+status: true
+dependencies:
+  module:
+    - node
+_core:
+  default_config_hash: 97tvtzGOa8_flb22CzSjgtm_YkiGMHvEBO-6q2K9V_U
+id: node_search
+label: Content
+path: node
+weight: -10
+plugin: node_search
+configuration:
+  rankings: {  }
--- a/web/modules/custom/dbcdk_community_content/dbcdk_community_content.info.yml
+++ b/web/modules/custom/dbcdk_community_content/dbcdk_community_content.info.yml
@@ -6,3 +6,4 @@ version: 8.x-1.0
 core: 8.x
 dependencies:
   - dbcdk_community
+  - views
--- a/web/modules/custom/dbcdk_community_content/dbcdk_community_content.info.yml
+++ b/web/modules/custom/dbcdk_community_content/dbcdk_community_content.info.yml
@@ -6,4 +6,5 @@ version: 8.x-1.0
 core: 8.x
 dependencies:
   - dbcdk_community
+  - search
   - views
--- a/web/config/sync/automated_cron.settings.yml
+++ b/web/config/sync/automated_cron.settings.yml
@@ -1,3 +1,3 @@
-interval: 86400
+interval: 3600
 _core:
   default_config_hash: fUksROt4FfkAU9BV4hV2XvhTBSS2nTNrZS4U7S-tKrs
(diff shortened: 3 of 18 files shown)


web/modules/custom/dbcdk_community_moderation/dbcdk_community_moderation.module
```diff
@@ -4,22 +4,6 @@
...
- */
-function dbcdk_community_moderation_user_login(User $account) {
-  // Redirect moderators to the flagged content list.
-  if (in_array('admin_moderator', $account->getRoles())) {
-    $url = new Url('page_manager.page_view_dbcdk_community_flagged_content');
-    $redirect = new RedirectResponse($url->toString());
-    $redirect->send();
-  }
-}
-
```
kasperg, line 22: Why is this included in the PR?

It does not seem related to the change at hand.

web/modules/custom/dbcdk_community/dbcdk_community.module
```diff
@@ -4,6 +4,35 @@
...
+    array_flip($account->getRoles())
+  );
+  $route = array_shift($user_role_routes);
+  if (!empty($route)) {
+    $url = new Url($route);
+    $redirect = new RedirectResponse($url->toString());
+    $redirect->send();
+  }
+}
+
```
kasperg: Why is this included in the PR?

It does not seem related to the change at hand.

kasperg:
Could we enable the Core Search module to support better content search?

web/config/sync/user.role.admin_library_employee.yml
```diff
@@ -0,0 +1,10 @@
+uuid: b598b954-c1e9-4a6f-a098-fafac21b0671
+langcode: en
+status: true
+dependencies: {  }
+id: admin_library_employee
+label: 'Library employee'
+weight: 5
+is_admin: null
+permissions:
+  - 'dbcdk community view reviews'
```
kasperg: Multiple unrelated files here as well.

web/modules/custom/dbcdk_community_moderation/dbcdk_community_moderation.module
kallehauge, line 22: Hmm, I shouldn't have touched anything in the moderation module so that's just a rebase (or something) gone wrong. I'll remove the removal :smile:

kallehauge:
Brilliant idea to take advantage of the `search` module! :+1: There is now added a search field instead of the title field that also searches in paragraphs. - There is no support for partial word searches though. I have tried to look at `search_api` but it is not yet ready for production sites. The newest alpha release fails multiple places (and I'm still not sure it it can do what we wish for or if it requires its own index-views like in D7, since I could barely install it).

web/modules/custom/dbcdk_community_content/dbcdk_community_content.info.yml
```diff
@@ -6,3 +6,4 @@ version: 8.x-1.0
 core: 8.x
 dependencies:
   - dbcdk_community
+  - views
```
kasperg, line 10: Do we depend on search here?

web/config/sync/search.page.node_search.yml
```diff
@@ -0,0 +1,15 @@
+uuid: 82431853-112f-46c2-a5bd-fd4f696ebc23
+langcode: en
+status: true
```
kasperg, line 3: Should the node search page be enabled? It is not used and we do not enable user search either.

kallehauge, line 3: It is not possible to disable the default search page (through the UI at least). So I disabled the User page - but I had to leave this one.

kasperg, line 3: :+1:

PR #102, opened 2016-05-26
Admin content should link to the community site.

kallehauge:
The ticket was to create a link from the content overview list to the community site.
I have chosen to go down the path of forwarding all "node/{nid}" requests to the community site since we don't handle any kind of front-end rendering and the existence of a "half view" of the node could cause confusion for some editors/librarians.
This solves the original issue by redirecting the already existing title-link to _"/awesome-content"_ to _"http://biblo.dk/awesome-content"_. (OBS: This does not effect REST requests).

**So this PR does the following:**
- Redirect all "node/{nid}" pages to the community site.
- Redirect users to the "admin/content" path after Node edits/creations (this was the default behaviour if you did not access the edit-page from the "admin/content" route and had a destination parameter in the query).
- Open the "Title" link on the "Content" view in a new window.

### DIFF
--- a/web/modules/custom/dbcdk_community_content/dbcdk_community_content.module
+++ b/web/modules/custom/dbcdk_community_content/dbcdk_community_content.module
@@ -6,6 +6,7 @@
 
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\Core\Url;
 use Drupal\node\Entity\Node;
 
 /**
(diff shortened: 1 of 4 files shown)


kasperg:
Does this work with #79 where they have requested that external urls open in a new window?

kallehauge:
Like we talked about on the phone: Server-side redirects cannot be opened in new windows.
The way I chose to work around this is to add a "target" property to the title-link on the content view. This will open the `node/{nid}` link in a new window and then the previous code will do its magic :smile:

kasperg:
Great

PR #107, opened 2016-06-03
Interface translation

kallehauge:
Add danish as an optional language. The language is detected based on the users language.


kasperg:
I have simplified the language negotiation to only allow Danish.

Please ensure that the install process either:
1. Works automatically as a part of importing the configuration
2. Involves manual work that we will then have to do as a part of the deployment process for this change

kallehauge:
Good job! The install process involves a manual update process. We could invest time to try and automate it in an update hook but I'm not sure its worth it. What do you think @kasperg ?

kasperg:
Automating it is not worth the time at the moment.

kasperg:
I will go ahead and merge this now.

PR #108, opened 2016-06-03
Scheduler

kallehauge:
- Add Scheduler module.
- Enable on Article content type.
- Change the following settings:
  1. Make content creation time match the scheduled publish time on Article nodes.
  2. Changed date format to `d-m-Y` instead of `Y-m-d` to match a danish format.

_**NB!**_ The view is part of the Scheduler module. It enables two paths:
1. A list to see users upcoming publications `/user/{uid}/scheduled`.
2. A list of all upcoming publications.


kasperg:
:+1: Good stuff!

PR #110, opened 2016-06-08
Hide deleted reviews

kallehauge:
This PR will:
- Update the Swagger Client
- Only fetch existing reviews in the reviews list
- Update tests to handle the new `markedAsDeleted` filter-condition

This PR solves the issue #104.


kasperg:
:+1:

PR #111, opened 2016-06-08
Add missing scheduler config

kallehauge:
Some missing Scheduler configuration that I missed in the original PR.


kasperg:
OK. It did not seem to cause any problems when I tested it out though.

